PR #4, opened 2016-01-06
Current task reminder script 

voter101:
Sometimes it's easy to get distracted during daily work. This plugin is supposed
to remind you about the task you are supposed to right now.

screen shot 2016-01-06 at 19 26 23

It stores the task in a `task` file in the script's directory. It also takes care about creating the file if it does not exist (however it doesn't care if you got write access ;) )

You can change the task, by calling the script with one argument like this:

`./current_task.1m.rb "Finish important business"`

### DIFF
--- a/Lifestyle/current_task.1m.rb
+++ b/Lifestyle/current_task.1m.rb
@@ -1,0 +1,24 @@
+#!/usr/bin/ruby
+
+# You need to call it with string argument to change a task like this:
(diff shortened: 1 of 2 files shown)


matryer:
I love this. I wonder how far away from a proper TODO list it actually is.

voter101:
It might be a good idea to do a integration with Wunderlist or Nozbe here. :)

PR #5, opened 2016-01-06
Weather Underground plugin

asnodgrass:
This PR adds a plugin to use the Weather Underground API to display current conditions and temperature in the menu bar, with additional information in the menu when clicked.  Weather conditions are shown using emoji (and are thus a bit limited).  An API key is required to use the plugin and can be obtained for free.  By default, location is set using geolocation (handled by WU); alternatively, a location can be given using the specifications of their API (e.g. "Raleigh, NC" or "35.78,-78.64").  Also, the units are SI by default, but can be changed to imperial if desired.

bitbar-wxu-plugin


matryer:
Wow - great work.

matryer:
Is 10s a sensible refresh time? How often does the weather change? :)

asnodgrass:
I have mine at 15m, but with the freebie API key, I think you could drive down as far as about 3 minutes before you'd hit your per-day limit. :D

asnodgrass:
Most of the big stations (e.g. airports) usually only refresh hourly, but personal weather stations are or can be much closer to real time.

matryer:
Why don't we set it at 30m by default?

asnodgrass:
Sure, I can do that.

asnodgrass:
If this isn't right, my apologies -- I haven't used rebase before.

matryer:
LGTM

asnodgrass:
Thanks!

PR #8, opened 2016-01-07
add DNS switcher plugin

saiqulhaq:
Hi
I just wrote DNS switcher plugin

screenshot copy

some sites are banned by my internet network provider 
I can't access them through Google DNS
so I often change my network DNS on Network Preferences

I have tried this blog post http://osxdaily.com/2013/07/01/change-network-location-quickly-mac-os-x/ for alternative, but it will disconnect and reconnect the connection, it's bad for me

### DIFF
--- a/Network/dnsswitcher.1d.sh
+++ b/Network/dnsswitcher.1d.sh
@@ -8,11 +8,6 @@
 
 # set your network service
 network_service="Wi-FI"
-tmp_dir="/tmp"
-dns_address_script_prefix="$tmp_dir/bitbar_dns_address_"
-
-switcher_script="bitbar_dns_switcher"
-switcher_script="$tmp_dir/$switcher_script"
 
 # add or remove list DNS below, then don't forget to make it enabled. see below
 google="8.8.8.8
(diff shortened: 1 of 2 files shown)


matryer:
What if the top line showed the current selection? Is that easy?

matryer:
p.s. awesome plugin idea :D

saiqulhaq:
Nice idea
I will try

matryer:
Let me know when you've done it and I'll merge it in.

saiqulhaq:
Hi @matryer
please review my last commit

screen shot

It will show the DNS IP address on the second line if the DNS is unrecognized by script 

and unfortunately we have to refresh the plugin manually to refresh displayed DNS (top line)
I have checked this pull request https://github.com/matryer/bitbar/pull/134
but it seems not working on my bitbar app

matryer:
Why does the plugin need to be refreshed manually? What version of BitBar are you running?

matryer:
p.s. I don't think `(DNS)` is needed - with the context, it's clear what it is I think. And it'll be neater. What do you think?

saiqulhaq:
> Why does the plugin need to be refreshed manually? What version of BitBar are you running?

I am using V1.4-beta.2

> p.s. I don't think (DNS) is needed - with the context, it's clear what it is I think. And it'll be neater. What do you think?

Yes you're right, it's more neater now

PR #13, opened 2016-01-07
added xcode-version plugin

hirschfl:
added xcode-version plugin


manojlds:
Would be useful if we can actually switch using `xcode-select`

hirschfl:
@manojlds Yes, sounds interesting! Feel free to add this functionality to my master branch.
For my use case this script is sufficient since I only want to make sure that I do not accidentally use the "wrong" Xcode version.

PR #15, opened 2016-01-07
Updated forecast.io plugin; more data, more fun

seripap:
Made forecast.io weather plugin a bit more snazzy: auto location lookup and more data.

ForecastIO Plugin


matryer:
that is snazzy!

matryer:
consider `dropdown=false` option too?

PR #17, opened 2016-01-07
Dev/Docker/docker-status: lint & fix bugs

keithamus:
- Switched the shebang to `bin/env sh` over `bash` for better
   compatibility with all POSIX environments.

 - Run file through shellcheck, which threw up some lint errors,
   which have been fixed.

 - `docker-machine env` has been passed the `--shell` flag, to
   specify the shell environment explicitly. Before it was relying
   on the user's shell being `bash` which is obviously not always
   the case.

### DIFF
--- a/Dev/Docker/docker-status.1m.sh
+++ b/Dev/Docker/docker-status.1m.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 #
 # Docker plugin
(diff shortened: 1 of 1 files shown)


matryer:
LGTM - thanks

manojlds:
@keithamus great, thanks for the modifications. Good to know about `shellcheck` too :)

PR #25, opened 2016-01-07
Dev/Github/notifications: initial implementation

keithamus:
This bitbar plugin fetches your GitHub notifications through the notifications API, and lists them as menu bar entries. If you have notificarions you get a blue dot, otherwise you get a dim grey dot. Each menu item is clickable - and takes you to the issue or pull request that the notification was for.

See screenshots:

No notifications!
screen shot 2016-01-07 at 22 59 40

Many notifications!
screen shot 2016-01-07 at 23 00 15

If you want to test what notifications look like, change like 12 (`url =...`) to have the query string `?all=true`.


matryer:
Oh this is extremely cool. Nice work.

PR #32, opened 2016-01-08
Add battery_cycles.sh

maxoumime:
Hi guys !

I added a new script showing the battery cycles number of your Mac laptop, only tested on a Macbook Pro Retina 13" 2015. 
I created it because I like to know where I am in the battery cycles.
I added a few emoticons to separate from other plugins if you are interested.

screencapture_batterycycles


matryer:
It would be cool to see some indication of what it is? Any ideas?

maxoumime:
I tried to find some unicodes that could represent the cycles of a battery, but that's not easy, I found those:
battery1
battery2
battery3

But they aren't very representative or good looking enough to me

matryer:
The 2nd one is nice and subtle - maybe go with that for now?

maxoumime:
Yeah sound ok to me ! 
Commited and ready

PR #39, opened 2016-01-08
Add simple todo list viewer and count

z0mbix:
Show list of todo items and a count using a very simple global todo list in ~/.todo.

### DIFF
--- a/Lifestyle/todo.30s.sh
+++ b/Lifestyle/todo.30s.sh
@@ -4,7 +4,7 @@
 #
 
 todo_file="$HOME/.todo"
-count=$(wc -l $todo_file | awk '{print $1}')
+count=$(wc -l "$todo_file" | awk '{print $1}')
 echo "Todos: $count"
 echo "---"
-cat $todo_file
+cat "$todo_file"
--- a/Lifestyle/todo.30s.sh
+++ b/Lifestyle/todo.30s.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+#
+# Track todos in a simple ~/.todo file
+#
+
+todo_file="$HOME/.todo"
+count=$(wc -l $todo_file | awk '{print $1}')
+echo "Todos: $count"
+echo "---"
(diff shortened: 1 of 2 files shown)


Lifestyle/todo.30s.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+#
+# Track todos in a simple ~/.todo file
+#
+
+todo_file="$HOME/.todo"
+count=$(wc -l $todo_file | awk '{print $1}')
```
keithamus:
```sh
In foo.sh line 7:
count=$(wc -l $todo_file | awk '{print $1}')
```

Before:
```sh
count=$(wc -l $todo_file | awk '{print $1}')
```

After:
```sh
count=$(wc -l "$todo_file" | awk '{print $1}')
```

Lifestyle/todo.30s.sh
```diff
@@ -0,0 +1,10 @@
...
+echo "Todos: $count"
+echo "---"
+cat $todo_file
```
keithamus:
```
In foo.sh line 10:
cat $todo_file
```

Before:
```sh
cat $todo_file
```

After:
```sh
cat "$todo_file"
```

keithamus:
@z0mbix the above two comments were linting errors that came up in [Shellcheck](http://www.shellcheck.net/) (`brew install shellcheck && shellcheck todo.30s.sh`)

It'd be good to fix these, because if my `$HOME` contains space characters then the script will get messed up and potentially do bad things, e.g. if `$HOME` is `/Users/my name` then L10 becomes `cat /Users/my name/todo` where `/Users/my` and `name/todo` are two separate args given to cat.

Simply quoting the $todo_file value will fix this.

matryer:
Is this one ready to go?

z0mbix:
Sorry, even though I have Syntastic installed and use it, it wasn't working on shell scripts so missed these. Fixed now :)

PR #40, opened 2016-01-08
Add Finance/rbc plugin

schmooser:
This plugin adds tracking of main Russian stock market indicators by parsing [Russian Business Consulting (RBC)](http://rbc.ru) agency website mainpage.

External tools [jq](https://stedolan.github.io/jq/), [pup](https://github.com/EricChiang/pup) and curl are required.

Example of plugin usage:
screen shot 2016-01-08 at 21 34 26

### DIFF
--- a/Finance/rbc.10m.sh
+++ b/Finance/rbc.10m.sh
@@ -10,7 +10,8 @@
 # They should be added in PATH
 
 export PATH=/usr/local/bin:~/Applications/go/bin:$PATH
-json=$(curl -s -L http://rbc.ru | pup '.indicators__group json{}') out="`echo $json | jq -r '.[] | .children | .[1:] | .[] as $c | "\($c.children | map(.text) | join(" ")) | color=\(if $c.children | .[0] | .text == "▲" then "green" else "red" end) | href=\($c.href)"'`"
+json=$(curl -s -L http://rbc.ru | pup '.indicators__group json{}')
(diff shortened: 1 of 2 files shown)


keithamus:
Hey @schmooser, before we merge this, would you be so kind as to go through [shellcheck](http://www.shellcheck.net/) and fix all of the outstanding lint issues. Here is a list that I get:

```sh
In rbc.10m.sh line 14:
out="`echo $json | jq -r '.[] | .children | .[1:] | .[] as $c | "\($c.children | map(.text) | join(" ")) | color=\(if $c.children | .[0] | .text == "▲" then "green" else "red" end) | href=\($c.href)"'`"
                         ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.
```

If you need any help with this let me know :smile:

keithamus:
Also it might be worth rewriting this as a Python script, if you have the time. That way it won't need any external dependencies (Python can manipulate JSON, and you could use RegExps for HTML) and we can keep the plugin codebase dependency free!

matryer:
What do we want to do with this plugin guys?

schmooser:
I don't know, if no one is interesting I might reject PR.

Don't have time to rewrite/rework, sorry.

keithamus:
The issues aren't showstoppers, we could merge as is and whenever @schmooser (or anyone else) wants to, they can refactor.

tresni:
The changes are minimal.  If @schmooser can't get to them I'll get them when I finish off the rest of the already committed ones :)

tresni:
Rebased and commited as 2facf9c66abefa03f2a483bf99ca9cdffd81ebb2 & 113ac1a95ed5d1a7cf841a243efb4939664b9be6.  Fixups (for metadata, linter & just getting the damn thing to work) are in 2175e8225e13d8280aafe53229c8209f070bf04a.

PR #42, opened 2016-01-08
World clock plugin

asnodgrass:
Display UTC in the menubar, and one or more additional zones in the drop down. The current format (HH:MM:SS) works best with a one second refresh, or alter the format and refresh rate to taste.

bitbar-worldclock-plugin


manojlds:
It would be cool if I can click one of the TZs, and switch the menu bar text to *that*!

asnodgrass:
That is an excellent idea! I'll see what I can do.

asnodgrass:
I'd need some way to maintain (and thus alter) state about the plugin to do that. For example, I could make each menu entry clickable, to execute another script that would alter the on-disk version of the plugin, or a configuration file.  I could sort of hack this today, but I'd have to use absolute paths since the plugin environment is pretty limited, and that will vary wildly from user to user.  Both matryer/bitbar#88 and matryer/bitbar#149 would be ideal to help address this.

In the interim, the user can of course modify the plugin however they like to change the zones, formats, refresh rate, etc.

keithamus:
matryer/bitbar#121 could also be a good candidate for that feature.

keithamus:
Hey @asnodgrass could you please add some [metadata](https://github.com/matryer/bitbar#metadata) to this plugin - then we can merge it!

asnodgrass:
Metadata added (and PR brought back up-to-date).

PR #43, opened 2016-01-08
Dev/Docker/docker-status: clickable entries

keithamus:
This commit gives each menu entry some clickable interactivity:

Docker machines which are on (and highlighted in green) - when
clicked - will be shut down (and subsequently all of their containers
will be stopped). If they are off (the ones highlighted in red) they
will be booted up.

The containers listed per machine now use tree like UTF-8 symbols to
better show their connection to their respective machine. In addition,
when clicked, they will be shut down. There is no concept of a
"non-running" container within this plugin, nor really with Docker, and
so there isn't an option to start containers up.

In addition to this, a small change has been made to the icon -
`dropdown=false` - to prevent the anchor icon showing up in the menus.

### DIFF
--- a/Dev/Docker/docker-status.1m.sh
+++ b/Dev/Docker/docker-status.1m.sh
@@ -5,9 +5,8 @@
 #
 # Displays the status of docker machines and running containers
 
-
 export PATH="/usr/local/bin:/usr/bin:$PATH"
-echo "⚓️"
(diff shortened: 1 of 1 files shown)


manojlds:
I was planning on doing something like this, and you beat me to it :) Looks great!

PR #45, opened 2016-01-09
Upgrade Mounted Disk capacity to Ejector

carlsonorozco:
Upgrade "Mounted Disk Capacity" Plugin to "Ejector"
- Able to eject each or all mounted disk easily


keithamus:
Hey @carlsonorozco, before we merge this, would you be so kind as to go through [shellcheck](http://www.shellcheck.net/) and fix all of the outstanding lint issues. Here is a list that I get:

```sh
In ejector.5s.sh line 12:
if [ $(find /Volumes -maxdepth 1 -type d | wc -l) = "1" ]; then
     ^-- SC2046: Quote this to prevent word splitting.


In ejector.5s.sh line 15:
    echo $about


In ejector.5s.sh line 20:
    status=$(diskutil eject /Volumes/$2 | sed -e 's/\/Volumes\///g')


In ejector.5s.sh line 47:
ls -p /Volumes | grep / | \
^-- SC2010: Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.


In ejector.5s.sh line 50:
    diskutil info /Volumes/$drive | grep -E 'File System Personality:|Volume Free Space:|Total Size:' | \


In ejector.5s.sh line 52:
        if [[ "$detail" =~ "Volume Free Space:" ]]
                           ^-- SC2076: Don't quote rhs of =~, it'll match literally rather than as a regex.


In ejector.5s.sh line 54:
            free_space=$(echo $detail | grep "Volume Free Space: " | cut -d '(' -f 1)


In ejector.5s.sh line 55:
        elif [[ "$detail" =~ "Total Size:" ]]
                             ^-- SC2076: Don't quote rhs of =~, it'll match literally rather than as a regex.


In ejector.5s.sh line 57:
            total_size=$(echo $detail | grep "Total Size:" | cut -d '(' -f 1)


In ejector.5s.sh line 73:
echo $about
```

Most of these are just a case of quoting variables to ensure that if they have spaces the command still works properly. If you need any help with this let me know :smile:

carlsonorozco:
I see, let me check of this. Thanks @keithamus

keithamus:
Hey @carlsonorozco could you please give this a rebase, and then I'll merge it!

carlsonorozco:
-> PR https://github.com/matryer/bitbar-plugins/pull/68 @keithamus thanks

PR #46, opened 2016-01-09
Dev/Travis/travis-check: refactor & remove travis

keithamus:
Refactor to remove the travis dependency, so that this script can
be run without installing any extra dependencies.

Also, as part of the refactor, remove the need for passing the
Github token, instead using Travis' own tokens.

In addition, add a few extra features:

 - repos_to_check is not required - if it is not supplied the script
   will go and fetch all repos under your account.
 - add INCLUDE_PULL_REQUESTS which will fetch build data about the
   latest pull requests. Off by default, so only commits are counted
 - add SECRET_FILTER_BY_STAR which is a hidden feature of the Travis
   api that lets you star repositories, and subsequently filter them.
   In essence behaving like `repos_to_check` but without maintaining
   the list inside of this script.

It'd be great to get @tomkinsc's approval before merging this - as @tomkinsc is the original author.

### DIFF
--- a/Dev/Travis/travis-check.2m.py
+++ b/Dev/Travis/travis-check.2m.py
@@ -4,78 +4,134 @@
 # Chris Tomkins-Tinch
 # github.com/tomkinsc
 
+# Keith Cirkel
+# github.com/keithamus
+
 # Dependencies:
-#   travispy (pip install travispy)
-#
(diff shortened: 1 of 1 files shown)


tomkinsc:
Sure, looks fine to me. It's nice to drop the TravisPy dependency. Thanks for the refactor!

PR #47, opened 2016-01-09
Added click action to upgrade homebrew packages

xalep:
Click on the upgrade count upgrades everything.
Click on a single package upgrades only the package.


keithamus:
Great stuff @xalep :smile:

PR #50, opened 2016-01-09
Add Twitch livestreamer plugin

gorangajic:
Because twitch.tv use flash it's better to use [livestreamer](http://docs.livestreamer.io/) to watch streams. This plugin lists top 10 streams by game and when clicked open in vlc using livestreamer

screen shot 2016-01-08 at 8 44 35 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ This repo contains scripts, programs and command-line tools that add functionali
 - StackOverflow
 - HackerNews Top Stories
 - Reddit, Top/New Links in Subreddits
+- Twitch, list top streams by game and launch with [livestreamer](http://docs.livestreamer.io)
 
 ####Weather
 - forecast.io
--- a/Web/Twitch/play-with-livestreamer.5m.js
+++ b/Web/Twitch/play-with-livestreamer.5m.js
@@ -1,0 +1,42 @@
+#!/usr/bin/env /usr/local/bin/node
+
+/*
+ * author:
+ *     Goran Gajic - https://github.com/gorangajic
+ * screenshot:
+ *     http://i.imgur.com/XEQQaxC.png
+ * desc:
+ *     list top 10 streams for Dota 2
+ *     you can modify the game by changing GAME variable
+ *     and limit by changing LIMIT variable
+ *     and place it in AUTH_TOKEN varibale
+ */
+
+var GAME = "Dota 2";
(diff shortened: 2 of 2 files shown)


keithamus:
Hey @gorangajic this looks good! 

@matryer I think this is a good time to bring up the point that we should look to add a `CONTRIBUTING.md` for this repo, enforcing some rules around what can be uploaded. The reason I say this is because a default Mac OSX install includes Python, Ruby and Bash - but not Node. I think the bitbar-plugins repo should only contain plugins that would work with an out-of-the-box experience, and not rely on the user installing dependencies or other runtimes. Thoughts?

gorangajic:
yup I agree with you. I have created this for me and want to share so anyone can reuse this if they find it useful.

I can probably figure out how to write this in Python or Ruby, but will probably take more time than writing  it in Javascript so that's why I used Node.js.

keithamus:
@gorangajic I wouldn't expect you to rewrite it (unless you really want to) - but I can refactor it if need be. But I think let's see what @matryer has to say before we go any further.

matryer:
I know what you mean - there's something very nice about just being able to drop the plugins into a folder and have it run. However, I'd hate to lose awesome plugins just because they have external dependencies. Some plugins will require API keys, others will have more significant dependencies. Perhaps the new [BETA metadata](https://github.com/matryer/bitbar-plugins/issues/53) stuff I'm playing around with is a good place to list dependencies?

keithamus:
@matryer :ok_hand: having some way to enumerate what dependencies are needed would be a great idea.

matryer:
... keeping them dependency free is also nice!

Derimagia:
I'm interested in creating some plugins in node and would love to see some other plugins in node.

I understand keeping everything dependency free but it would be nice to list them and just make sure it's known that they require node.

I also just noticed that @sindresorhus just created a Bitbar node client at https://github.com/sindresorhus/bitbar which I'm inclined to take advantage of.

Derimagia:
We should probably move the Dependency conversation to another thread.

To get back on topic about this Plugin, @gorangajic I like the idea of the plugin but it looks like it's specifically styled for Dark Mode. It looks a bit unreadable.

http://i.imgur.com/OG4JdIX.png

gorangajic:
ups @Derimagia, updated commit that does not set color to white

PR #51, opened 2016-01-09
Dev/Homebrew/brew-updates: lint & optimise

keithamus:
Prevent the update count appearing in the dropdown, and instead
offer a clearer "Update all" option.

Switch from `` to $() to appease shellcheck


manojlds:
Should we use "Upgrade all" instead of "Update all" to be in sync with the brew sub-commands?

keithamus:
Good point. Done.

Issue #53, opened 2016-01-09
Plugin browser


matryer:
Guys,

I put together a very simple BitBar plugin browser: http://getbitbar.appspot.com (will be moved to getbitbar.com when it's ready). Real examples in the Tutorial section: http://getbitbar.appspot.com/plugins/Tutorial (and the source https://github.com/matryer/bitbar-plugins/blob/master/Tutorial/cycle_text_and_detail.sh)

By including metadata like the below in the plugin *source file*, authors can customise the browser:

```
# <bitbar.title>Cycle text and detail text</bitbar.title>
# <bitbar.version>1.0</bitbar.version>
# <bitbar.author>Mat Ryer</bitbar.author>
# <bitbar.author.github>matryer</bitbar.author.github>
# <bitbar.desc>Text above --- will be cycled through in the menu bar, whereas text underneath will be visible only when you open the menu.</bitbar.desc>
# <bitbar.image>http://www.images.com/path/to/image</bitbar.image>
```

It matches using regex, so tags may appear anywhere in the file (commented out however is suitable):

```
`<bitbar\.` + tag + `>(.*?)</bitbar\.` + tag + `>`
```

Would love some feedback, comments, bug reports, etc. over at https://github.com/matryer/getbitbar.com

keithamus:
:+1: sounds great :smile:

thameera:
A Search box would be handy.

Also, the link to the plugin browser's repo doesn't seem to work. (maybe private repo?)

matryer:
Which link to plugins repo? It should link to THIS repo

thameera:
I was referring to the link at the end of your comment: https://github.com/matryer/getbitbar.com

maxoumime:
This is a really good idea, I like it !
And same here, I don't have access to your repo: https://github.com/matryer/getbitbar.com

matryer:
Ahh yes - it’s public now

matryer:
Plugins with metadata will now float to the top of the listings: https://getbitbar.com/plugins/System

manojlds:
This is not accessible? - https://github.com/matryer/getbitbar.com

PR #55, opened 2016-01-09
Update battery_cycles, add battery condition and max cycles

maxoumime:
Hi

I modified my script a bit, I added a refresh rate at 6 hours, not easy to find out because this can change with the un/plugging of the battery but 6 hours seems okay to me.
I also added a few indicators in the dropdown menu like the number of cycles on the maximum cycles number given by Apple (1000 for almost all the Macbooks from 2009), the color changes depending on the number of cycles (green < 750, orange < 1000, red > 1000). 
I wanted this maximum number to change depending on the model and year but despite the model is easy to find, the year isn't directly. You have to send the model number to a webservice, and I didn't want to send those informations (I added a href linking to the Apple section dedicated to batteries).

https://support.apple.com/en-us/HT201585

I also display the battery condition returned by the OS, with the color changing depending on that condition.

battery


matryer:
LGTM - do you want to be the first to implement the Metadata for plugins to have your Plugin appear nicely here: http://getbitbar.appspot.com/plugins/System ?  See https://github.com/matryer/bitbar-plugins/issues/53

maxoumime:
Wow sure that's a cool thing ! I add it right away

maxoumime:
Aaaaand done :) Let's have a look

matryer:
Looks great.

screen shot 2016-01-09 at 18 01 11

matryer:
Plugins with metadata will appear closer to the top - makes the site more engaging.

maxoumime:
Yeah it looks great ! 
And yeah that's a good point to order like that, more attractive indeed

matryer:
see https://getbitbar.com/plugins/System for live version

PR #57, opened 2016-01-09
Add SSH plugin

thameera:
screen shot 2016-01-10 at 12 08 18 am

I added this under Network section, but not entirely sure if that's the right place to put that in.

### DIFF
--- a/Network/ssh.sh
+++ b/Network/ssh.sh
@@ -1,0 +1,15 @@
+!/usr/bin/env bash
+#
+# Quickly SSH to your favorite hosts
+# The list of hosts are extracted from ~/.ssh/config
+#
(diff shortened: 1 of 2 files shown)


manojlds:
This should probably be in `Dev`?

C4Wiz:
how do you format the config file?

thameera:
@C4Wiz Like a regular SSH config file. Eg: https://nerderati.com/2011/03/17/simplify-your-life-with-an-ssh-config-file/

C4Wiz:
ok thanks. i got it all cofigured generated keys. now when i use the plugin to connect to the server it asks for the passphrase for the key i generated, if i still have to type a password, whats the point?

C4Wiz:
got it thanks

PR #58, opened 2016-01-09
Spellcheck & Python lint

maxoumime:
Was wandering on the other PRs and I found a post from @keithamus about ShellCheck. Thought I should've test my scripts before opening my PRS.
So I tested my battery-cycles Bash script and my currency-tracker Python script through online lint testers ( www.shellcheck.net and www.pep8online.com ) and corrected them.

That's not a **big** fix, but I thinks that's necessary.


keithamus:
Woop! Thanks @maxoumime

PR #60, opened 2016-01-09
Add metadata to iTunes/Spotify plugins

daturkel:
Pursuant to #53, added metadata for the Spotify and iTunes plugins. Could only add one Github author for the iTunes plugin (or at least I suspected) so I put myself, but it should be noted that @jtokoph wrote the Spotify plugin and all I did was swap out "Spotify" for "iTunes" a bunch of times.

Also switched the executable bit for Spotify so no one has to `chmod +x` it.


matryer:
They're live: https://getbitbar.com/plugins/Music

PR #72, opened 2016-01-10
added contributor name

gautamkrishnar:
https://github.com/matryer/bitbar/pull/54


gautamkrishnar:
@matryer my plugin require both "memusage.5s.sh" and "memusage.py" file, but in the new repository i can see only "memusage.py"

gautamkrishnar:
@matryer

PR #74, opened 2016-01-10
Added python resolve using env

FredrikL:


matryer:
Thanks for this - I wonder if you could add the metadata too? https://github.com/matryer/bitbar#metadata

FredrikL:
Done :)

PR #77, opened 2016-01-10
Metadata vpn checker

pldubouilh:
Added metadata for getbitbar.com


keithamus:
Seems like this is more than just metadata? Also if `utun0` doesn't exist then `ifconfig` will output to `stderr` which isn't great. @pldubouilh could you take another look at maybe remove the ifconfig line or improve on it? Thanks

pldubouilh:
Right, I made that `ifconfig` command to output to `/dev/null` if `utun0` is not existing. If it is, it'll print the ip address on the menu. OSX always seems to give me that `utun0` interface, even when I explicitly ask for another interface in my openvpn conf file...

I also modified the commented `killall` line, so it's not printing anything.

PR #78, opened 2016-01-10
iTunes Rating script

sw82:
Script to show the current rating of the current track which also allows you to rate the it.


keithamus:
Hey @sw82 there seems to be nothing inside these commits! I think this PR may have bugged out or something. Do you mind raising it again?

PR #82, opened 2016-01-11
remove color=white from semaphoreci plugin

gorangajic:
this is needed so it can work on white osx theme also


manojlds:
Shouldn't the right fix be to handle dark and white using the `$BitBarDarkMode `

gorangajic:
it looks fine like this

PR #95, opened 2016-01-11
Add clickable links to Weather Underground plugin

asnodgrass:
Adds three menu items that open web links:

* Station forecast page
* Station historical data page
* Google maps for station location


asnodgrass:
I erroneously committed before resyncing from upstream.  I will retry later.

PR #99, opened 2016-01-12
Adding Games directory and game-related plugin

astrostl:


Games/baj.2m.sh
```diff
@@ -0,0 +1,29 @@
...
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Justin Honold</bitbar.author>
+# <bitbar.author.github>astrostl</bitbar.author.github>
+# <bitbar.desc>Checks for active turns at http://boiteajeux.net every two minutes</bitbar.desc>
+# <bitbar.image></bitbar.image>
+# <bitbar.dependencies></bitbar.dependencies>
+
+# INSTRUCTIONS: set baj_login=LOGIN and baj_password=PASSWORD in ~/.bitbar
+
+. ~/.bitbar
```
keithamus: I'm concerned that this:

 - Doesn't follow convention. All other scripts set variables within the script and need to be edited
 - May break if BitBar gets a feature like https://github.com/matryer/bitbar/issues/88

astrostl: Responded in that thread. I think variables-in-scripts is an antipattern, especially for GitHub projects :)

Very happy to refactor into any env or config file system that supports out-of-repo key/value storage!

keithamus: I agree that variables in scripts are an anti pattern - hopefully matryer/bitbar#88 will solve this. Until then, I think we should avoid doing things which might conflict with future progress, and stick to a convention, even if it is not the most desirable one.

astrostl: Updated :-/

Games/baj.2m.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# <bitbar.title>BAJ</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Justin Honold</bitbar.author>
+# <bitbar.author.github>astrostl</bitbar.author.github>
+# <bitbar.desc>Checks for active turns at http://boiteajeux.net every two minutes</bitbar.desc>
+# <bitbar.image></bitbar.image>
```
keithamus: Hi @astrostl, one last thing. Could you please add an image to this, that'd be awesome!

matryer: Images really improve the https://getbitbar.com site.

astrostl: Advice on the "appropriate" way to host images on GH?

astrostl: (this would be a good addition to the plugin authorship guide)

flesch: I upload an image to a new issue and use the URL it returns. The image link is still good even without submitting the issue.

dorzlbuq1j

Issue #98, opened 2016-01-12
re: Weather


rdcosta:
FAO: (@seripap)

I get the following error when I try and use your BitBar plugin
  File "weather.15m.py", line 94
    'city': str(location["city"]),
         ^
SyntaxError: invalid syntax

Thanks

pvledoux:
Same problem here.

matryer:
Does anyone know how to fix this?

keithamus:
I suppose you're using the ForecastIO one? It looks like [this line](https://github.com/matryer/bitbar-plugins/blob/21773f78eb0aa1327d1387014540c37b5b6f6ddb/Weather/ForecastIO/weather.15m.py#L92) is missing a trailing comma.

keithamus:
That was my bad, I merged in this PR which caused it: https://github.com/matryer/bitbar-plugins/pull/80

seripap:
#103 should resolve

rdcosta:
Thanks for the quick response - that fixed it

PR #102, opened 2016-01-12
Weather/ForecastIO/weather: pyfmt & fix syntax errors

keithamus:
This fixes the syntax errors for weather.15m.py as mentioned in https://github.com/matryer/bitbar-plugins/issues/98. Also I ran `pyfmt` against it to make sure its pep8. No functional changes.


seripap:
I've opened up #103 - should fix the issue and added another feature.

PR #104, opened 2016-01-12
really works

alexandregz:
Previous version (perl script) doesn't works fine. This really works (explained and tested here #84 )


keithamus:
I think this file needs to be renamed to `.rb` now, as its not a shell script file. Could you please raise a PR @alexandregz?

alexandregz:
You're right, sorry!

PR #109, opened 2016-01-12
mv keyboard-system_profiler.1m.sh to keyboard-system_profiler.1m.rb

alexandregz:


keithamus:
:+1: thanks @alexandregz

PR #117, opened 2016-01-13
correction

alexandregz:


matryer:
I might have merged them in the wrong order - can you see?

alexandregz:
No problem, there are independents. Maybe you have had a little problem because I don't put the previous commit here and the forked repo don't have the PR #116 but looks fine.

thx!

PR #122, opened 2016-01-13
Add GitHub:Enterprise support to GitHub/notifications.30s.py

flesch:
I added support for fetching **GitHub:Enterprise** notifications. For those of us that use both **GitHub.com** and an enterprise instance, notifications from both places will show up (rather than running a second instance of this plugin with GH:E modifications).

preview

If a user doesn't define a GH:E token and API, that section won't show up.

### DIFF
--- a/Dev/GitHub/notifications.30s.py
+++ b/Dev/GitHub/notifications.30s.py
@@ -1,30 +1,76 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Github Notifications by Keith Cirkel
-# Fetch Github notifications in your menu bar!
+
+# <bitbar.title>GitHub Notifications</bitbar.title>
+# <bitbar.version>v2.0.0</bitbar.version>
+# <bitbar.author>Keith Cirkel, John Flesch</bitbar.author>
+# <bitbar.author.github>flesch</bitbar.author.github>
+# <bitbar.desc>GitHub (and GitHub:Enterprise) notifications in your menu bar!</bitbar.desc>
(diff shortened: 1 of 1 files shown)


Dev/GitHub/notifications.30s.py
```diff
@@ -1,30 +1,71 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Github Notifications by Keith Cirkel
-# Fetch Github notifications in your menu bar!
+
+# <bitbar.title>GitHub Notifications</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
```
manojlds: author metadata?

flesch: What's the recommendation for a plugin with multiple authors (especially one with a change so drastic since @keithamus first wrote it)? I didn't want to assume Keith still wants to be on the hook to support a plugin that now looks much different.

matryer: Only one author is currently supported on https://getbitbar.com/ - should we add support for comma separated authors?

keithamus: @flesch I'm fine if you want to change the author to just you.

matryer: Just have one in the `github` tag, but include both names (comma separated) in `author` as it's display only. That work?

flesch:
I updated the author metadata and squashed everything. I fixed an issue where it would incorrectly say "1 notifications" (like above) - is it worth updating the screenshot?

keithamus:
:+1: all good :smile:

flesch:
Thanks!

PR #131, opened 2016-01-14
Make teller-track-spending executable and give human readable name

### LINKED ISSUE #129: re: teller-track-spending
FAO: (@sebinsua)

Can you changes the permissions on the file to make it executable and submit an update? Also, please rename in metadata to nice human readable name. :)

sebinsua:
Should fix #129.

### DIFF
--- a/Finance/teller-track-spending.1h.sh
+++ b/Finance/teller-track-spending.1h.sh
@@ -7,11 +7,11 @@
 # - teller-cli: https://github.com/sebinsua/teller-cli#from-release
 # - pcregrep: `brew install pcre`
 #
-# <bitbar.title>teller-track-spending</bitbar.title>
+# <bitbar.title>Track spending at UK banks with Teller</bitbar.title>
 # <bitbar.version>v1.3.1</bitbar.version>
 # <bitbar.author>Seb Insua</bitbar.author>
 # <bitbar.author.github>sebinsua</bitbar.author.github>
-# <bitbar.desc>Track your spending from your menu bar</bitbar.desc>
+# <bitbar.desc>Track your spending with teller-cli and an account at Teller.io</bitbar.desc>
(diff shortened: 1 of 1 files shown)


Finance/teller-track-spending.1h.sh
```diff
@@ -7,7 +7,7 @@
 # - teller-cli: https://github.com/sebinsua/teller-cli#from-release
 # - pcregrep: `brew install pcre`
 #
-# <bitbar.title>teller-track-spending</bitbar.title>
+# <bitbar.title>Track spending at UK banks</bitbar.title>
```
keithamus: Perhaps this should still mention Teller?

sebinsua: I can do that!

I wasn't sure how long it could be to be honest - so I was erring on side of caution.

sebinsua: Can I put a hyperlink in the description actually? Might be a good idea to draw attention to signup.

keithamus: We already have a precedent of some fairly long descriptions, see [the pull-requests script](https://github.com/matryer/bitbar-plugins/blob/master/Dev/GitHub/pull-requests.5m.sh#L7) :smile: .

sebinsua: Wow a blob of JavaScript inside a shell script, too. hahaha.

keithamus: Actually I just realised you don't have a description! Perhaps the title could be `Teller Spending Tracker` and the desc could be `Track your spending with the teller-cli and the Teller banking api`, or something similar?

sebinsua:
Updated RE: your comments.

keithamus:
:hand: :+1: :ok_hand:

PR #133, opened 2016-01-14
change bits

aaronedell:
changed folder name to environment and also add emotion and change preview image


matryer:
I already did this.

PR #148, opened 2016-01-17
Updated docker plugin 

pradeepchhetri:
- Added some emojis to display
- Added exited containers also to the list.
screen shot 2016-01-17 at 21 20 14

### DIFF
--- a/Dev/Docker/docker-status.1m.sh
+++ b/Dev/Docker/docker-status.1m.sh
@@ -26,7 +26,7 @@ fi
 echo "${DOCKER_MACHINES}" | while read -r machine; do
   STATUS=$(docker-machine status "$machine")
   if [ "$STATUS" = "Running" ]; then
-    echo "$machine | color=green bash=$(which docker-machine) param1=stop param2=$machine terminal=false refresh=true"
+    echo "🔵  $machine | bash=$(which docker-machine) param1=stop param2=$machine terminal=false refresh=true"
     ENV=$(docker-machine env --shell sh "$machine")
     eval "$ENV"
     CONTAINERS="$(docker ps --format "{{.Names}} ({{.Image}})|{{.ID}}")"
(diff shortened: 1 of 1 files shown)


manojlds:
@pradeepchhetri The initial version did have emojis like what you have done. @keithamus made this change https://github.com/matryer/bitbar-plugins/commit/5943c7f03c6c4b64cebcae739bcd4375cd8cca57 which is what we have right now. I started liking that one much more :) I guess the emojis for the containers are a bit too much. Also, initially wanted to have the 🔴 emoji too, but didn't go ahead with it due to lack of a green one!

@keithamus what do you think? Maybe we can have variables to turn on/off the emojis. Also maybe turn on/off display of dead containers.

PS: @pradeepchhetri long time :)

keithamus:
I think dead containers is probably a bit much - as when I list dead containers I get ones from weeks ago.  Is their an upper limit to this?

Otherwise the addition of emoji are fine by me. Having an option to turn them on/off (defaulting to on) might be an idea though.

Dev/Docker/docker-status.1m.sh
```diff
@@ -26,24 +26,28 @@ fi
...
-        if [ "$CONTAINER" = "$LAST_CONTAINER" ]; then SYM="└"; fi
-        echo "$SYM $CONTAINER_NAME | color=green bash=$(which docker) param1=stop param2=$CONTAINER_ID terminal=false refresh=true"
+        CONTAINER_NAME=$(echo "$CONTAINER" | awk -F"|" '{print $1}')
+        CONTAINER_ID=$(echo "$CONTAINER" | awk -F"|" '{print $2}')
+        CONTAINER_STATE=$(echo "$CONTAINER" | awk -F"|" '{print $3}')
+        SYM="├ 💻 "
+        if [ "$CONTAINER" = "$LAST_CONTAINER" ]; then SYM="└ 💻 "; fi
+        case "$CONTAINER_STATE" in
+          *Up*) echo "$SYM $CONTAINER_NAME | color=green bash=$(which docker) param1=stop param2=$CONTAINER_ID terminal=false refresh=true";;
+          *Exited*) echo "$SYM $CONTAINER_NAME | color=red bash=$(which docker) param1=stop param2=$CONTAINER_ID terminal=false refresh=true";;
```
manojlds: @pradeepchhetri should this try to *restart* the exited container?

Issue #151, opened 2016-01-19
[Bug] - OpenWeatherMap plugin shows exception in status bar if unable to connect


ryansydnor:
Encountered this one while not connected to the internet. Looks like it'd be easy to catch a URLError and handle it gracefully.

ryansydnor:
PR merged. Closing issue.

PR #153, opened 2016-01-19
Gracefully handle inability to connect to openweathermap.org

ryansydnor:
This addresses #151


ryansydnor:
@seripap - take a look!

seripap:
I would just change URLError to urllib2.HTTPError and remove the module extraction- other than that looks good :+1:

ryansydnor:
@seripap - fixed! thanks for the suggestion.

PR #154, opened 2016-01-19
Add support for multiple users

falquan:
Include $USER env var in tmp_dir path to allow for a history folder per
user.


keithamus:
:+1:

PR #163, opened 2016-01-21
added caffeinate plugin with some sugar

citoki:
This is the bitbar plugin to execute caffeinate program.


matryer:
An image would be awesome if you can be bothered? People have been using http://imgur.com.

matryer:
Thanks.

PR #164, opened 2016-01-21
Add Bamboo build status plugin v1

njovin:
I've written and am using this plugin which displays the build status for a Bamboo build.  I thought others might find it useful.


matryer:
Thank you.

PR #172, opened 2016-01-23
updated shebangs

sbdchd:
I updated the shebangs for primarily python files to use the more portable #!/usr/bin/env & the corresponding python version.

### DIFF
--- a/Bitcoin/huobi.com/huobiprice.1s.py
+++ b/Bitcoin/huobi.com/huobiprice.1s.py
@@ -1,5 +1,5 @@
-#!/usr/bin/python
-# coding=utf-8
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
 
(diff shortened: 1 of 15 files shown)


keithamus:
Thank you so much! I was going to get round to this so thank you :smile:

sbdchd:
I think this needs to be reverted as BitBar is unable to find python2 or python3 since its environment is different.

`#!/usr/bin/env python` seems to work but `#!/usr/bin/env python2` and `#!/usr/bin/env python3` don't.

The solution I have found is to copy the advice for javascript and do the following.
```python
#!/usr/bin/env /usr/local/bin/python2
```
and
```python
#!/usr/bin/env /usr/local/bin/python3`
```

keithamus:
`#!/usr/bin/env /usr/local/bin/python3` is kind of pointless as the point of env is that it provides PATH. But agreed, I'll revert for now, my bad.

PR #173, opened 2016-01-24
Ethereum price plugin

Har01d:
Ethereum price

This plugin shows the last price for Ethereum in USD.

### DIFF
--- a/Ethereum/ethusd.1m.sh
+++ b/Ethereum/ethusd.1m.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Shows last Ethereum price in USD.
+#
+# <bitbar.title>Ethereum last price plugin</bitbar.title>
+# <bitbar.version>0.1A</bitbar.version>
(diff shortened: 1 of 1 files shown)


Har01d:
It's also very easy to change Ethereum to any other cryptocurrency from this list: https://coinmarketcap-nexuist.rhcloud.com/api/all (e.g. Litecoin, Dash, Dogecoin, etc.)

Just change `eth` in `echo -n "𝚵"; curl -s "https://coinmarketcap-nexuist.rhcloud.com/api/eth/price" | egrep -o '"usd": "[0-9]+(\.)?([0-9]{2})?' | sed 's/"usd": / /' | sed 's/\"//g'` into necessary cryptocurrency symbol.

It's also possible to change `usd` into `eur`, `cny`, `cad`, `rub` or `btc`.
So maybe it's better to call it a "Cryptocurrency price plugin".

keithamus:
It would be great to turn the `"usd"` part into a variable, to allow users to configure it. Defaulting it to "usd" would be ideal.

PR #176, opened 2016-01-25
Fix caffeinate plugin and simplified options

citoki:
Simplified the options for program launch and concentrated on one task the user wants to do. Introduced an icon when terminal command is executed, to give the user an indicator in the menu bar.


keithamus:
Looks good, but it needs a rebase please :smile:

PR #177, opened 2016-01-25
Add Pomodoro timer

koozz:
A minimalistic Pomodoro timer.

### DIFF
--- a/Lifestyle/pomodoro.1s.sh
+++ b/Lifestyle/pomodoro.1s.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# <bitbar.title>Pomodoro timer</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Jan van den Berg</bitbar.author>
+# <bitbar.author.github>koozz</bitbar.author.github>
+# <bitbar.desc>Provides a minimal Pomodoro timer</bitbar.desc>
+# <bitbar.dependencies>bash</bitbar.dependencies>
+
+
+# Icon; http://emojipedia.org/tomato
+POMODORO="🍅"
+#POMODORO="\xF0\x9F\x8D\x85"  # echo -e $POMODORO
+POMODORO_SECONDS=1500
+POMODORO_DATA="$(dirname $0)/pomodoro.data"
+
+# Ensure data file
+[[ ! -f $POMODORO_DATA ]] && touch $POMODORO_DATA && echo "POMODORO_END=" > $POMODORO_DATA
+
+# Load data file
+source $POMODORO_DATA
+
+# Pomodoro menu
+pomodoro-menu () {
+  if [[ -z $POMODORO_END ]]; then
+    echo "$POMODORO"
+    echo "---"
+    echo "Start pomodoro | terminal=false bash=$0 param1=start"
+  else
+    local now=$(date +"%s")
+    local diff=$(($POMODORO_END-$now))
+    if [[ $diff < 0 ]]; then
+      echo "$POMODORO 0:00 | color=red"
+      echo "---"
+      echo "Clear | terminal=false bash=$0 param1=stop"
+    else
+      printf "%s %d:%.2d\n" $POMODORO $(($diff / 60)) $(($diff % 60))
+      echo "---"
+      echo "Stop pomodoro | terminal=false bash=$0 param1=stop"
+    fi
+  fi
+}
+
+# Starting a new pomodoro (extend with cylces, break-time etc.)
+pomodoro-start () {
+  local now=$(date +"%s")
+  local end=$(( $now + $POMODORO_SECONDS ))
+  (
+    echo "POMODORO_END=$end"
+  ) > $POMODORO_DATA
+}
+
+# Stopping a pomodoro
+pomodoro-stop () {
+  (
+    echo "POMODORO_END="
+  ) > $POMODORO_DATA
+}
+
+# Main
+case "$1" in
+"start")
+  pomodoro-start
+  ;;
+"stop" )
+  pomodoro-stop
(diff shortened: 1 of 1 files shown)


Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+# <bitbar.author.github>koozz</bitbar.author.github>
+# <bitbar.desc>Provides a minimal Pomodoro timer</bitbar.desc>
+# <bitbar.dependencies>bash</bitbar.dependencies>
+
+
+# Icon; http://emojipedia.org/tomato
+POMODORO="🍅"
+#POMODORO="\xF0\x9F\x8D\x85"  # echo -e $POMODORO
+POMODORO_SECONDS=1500
+POMODORO_DATA="$(dirname $0)/pomodoro.data"
```
keithamus: The $0 needs to be quoted otherwise if the path contains spaces then things could go wrong.

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Ensure data file
+[[ ! -f $POMODORO_DATA ]] && touch $POMODORO_DATA && echo "POMODORO_END=" > $POMODORO_DATA
```
keithamus: `$POMODORO_DATA` needs quoting here too - all three instances.

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Load data file
+source $POMODORO_DATA
```
keithamus: Double quotes here too please.

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+      echo "---"
+      echo "Stop pomodoro | terminal=false bash=$0 param1=stop"
+    fi
+  fi
+}
+
+# Starting a new pomodoro (extend with cylces, break-time etc.)
+pomodoro-start () {
+  local now=$(date +"%s")
+  local end=$(( $now + $POMODORO_SECONDS ))
```
keithamus: Ideally you should declare and set separately, read more [on the shellcheck SC2155 rule here](https://github.com/koalaman/shellcheck/wiki/SC2155)

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Pomodoro menu
+pomodoro-menu () {
+  if [[ -z $POMODORO_END ]]; then
+    echo "$POMODORO"
+    echo "---"
+    echo "Start pomodoro | terminal=false bash=$0 param1=start"
+  else
+    local now=$(date +"%s")
+    local diff=$(($POMODORO_END-$now))
```
keithamus: Ideally you should declare and set separately, read more [on the shellcheck SC2155 rule here](https://github.com/koalaman/shellcheck/wiki/SC2155)

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+    if [[ $diff < 0 ]]; then
```
keithamus: This should be `-lt`, as `<` is for string comparison. See [SC2071 for more](https://github.com/koalaman/shellcheck/wiki/SC2071)

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+  fi
+}
+
+# Starting a new pomodoro (extend with cylces, break-time etc.)
+pomodoro-start () {
+  local now=$(date +"%s")
+  local end=$(( $now + $POMODORO_SECONDS ))
+  (
+    echo "POMODORO_END=$end"
+  ) > $POMODORO_DATA
```
keithamus: Double quote this please.

Lifestyle/pomodoro.1s.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+# Stopping a pomodoro
+pomodoro-stop () {
+  (
+    echo "POMODORO_END="
+  ) > $POMODORO_DATA
```
keithamus: And again, double quotes here please.

keithamus:
Hey @koozz awesome plugin! I've made some notes about small changes - mostly syntactical, by using [shellcheck](http://www.shellcheck.net/) to lint the file. If you could please correct these then I'll happily merge :smile:

koozz:
There are 3 messages from spellcheck that I cannot seem to fix.

keithamus:
SC2155 just means you should declare first, then assign after. For example:

```sh
local now;
now=$(date +"%s")
```

As for SC1090, you just need to comment the line above; so:

```sh
# shellcheck source=~/.bitbar/plugins/pomodoro.data
source "$POMODORO_DATA"
```

Or alternatively you can disable the check:

```sh
# shellcheck disable=SC1090
source "$POMODORO_DATA"
```

matryer:
see https://github.com/matryer/bitbar-plugins/issues/180

keithamus:
Hey @koozz, thanks very much for working on this - and thanks also for being responsive to my comments and making the right fixes.

Another user (@gorangajic) has also made a Pomodoro plugin - around the same time as yours. We've gone with theirs, and so I'm going to close this one. Please don't take this as a sign that we don't value your work - we really do! Please don't let this put you off from making PRs in the future! We'd love to see some more great code from you :smile:.

matryer:
Here, here.

koozz:
Sorry, I didn't know about the other work in progress.
Just fiddled around and made this one. It's pretty minimalistic, but I'll keep it around for my own use :)

PR #184, opened 2016-01-25
Revert shebang update

sbdchd:
Reverts matryer/bitbar-plugins#172

The previous change did not take into account the different environment of BitBar.

See [this comment](https://github.com/matryer/bitbar-plugins/pull/172#issuecomment-174638832).


keithamus:
:blush: thanks :smile:

PR #191, opened 2016-01-26
Gracefully handle inability to connect to ipinfo.io

ryansydnor:


ryansydnor:
@taylorzane - what do you think?

taylorzane:
Looks good to me. I was checking to see if there was a way to detect internet connection to guard before trying to do an HTTP request. But there doesn't seem to be anything other than using a shell command to use some Mac specific exe (`networksetup`, etc). I've made considerable updates to this plugin...I should PR the new stuff.

Issue #183, opened 2016-01-25
weather plugin stopped working because of HTTP Error 401: Unauthorized


schmooser:
I get this error while running the plugin:

    [~] % /Users/rbd/Projects/bitbar-plugins/Enabled/weather.15m.py
    Traceback (most recent call last):
      File "/Users/rbd/Projects/bitbar-plugins/Enabled/weather.15m.py", line 53, in <module>
        print render_wx()
      File "/Users/rbd/Projects/bitbar-plugins/Enabled/weather.15m.py", line 46, in render_wx
        weather_data = get_wx()
      File "/Users/rbd/Projects/bitbar-plugins/Enabled/weather.15m.py", line 24, in get_wx
        wx = json.load(urllib2.urlopen('http://api.openweathermap.org/data/2.5/weather?id=' + location + '&units=' + units + '&appid=' + api_key + "&v=" + str(randint(0,100))))
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 154, in urlopen
        return opener.open(url, data, timeout)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 437, in open
        response = meth(req, response)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 550, in http_response
        'http', request, response, code, msg, hdrs)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 475, in error
        return self._call_chain(*args)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 409, in _call_chain
        result = func(*args)
      File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 558, in http_error_default
        raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
      urllib2.HTTPError: HTTP Error 401: Unauthorized

seripap:
You need to change the API key to your own- it was shipped with a generic key that may not be active anymore. You can grab it at openweathermap.org, or just replace it with this one from the other plugin **8b4824b451d5db1612156837df880f55**

PR #166, opened 2016-01-22
Update GeoIPWeather

dkhamsing:
1. Old api key stopped working.. :confused: 
- Added metadata 

weather

PR #192, opened 2016-01-26
Replaced weather API key- uses same as other plugin

seripap:
For #183


keithamus:
:+1: :star2:

dkhamsing:
#166 :sunglasses:

PR #196, opened 2016-01-27
CircleCI BitBar plugin

mCzolko:
View your recent builds on CircleCI

### DIFF
--- a/Dev/CircleCI/.gitignore
+++ b/Dev/CircleCI/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/Dev/CircleCI/README.md
+++ b/Dev/CircleCI/README.md
@@ -1,0 +1,1 @@
+Run `npm install` within this folder
--- a/Dev/CircleCI/README.md
+++ b/Dev/CircleCI/README.md
@@ -1,1 +1,3 @@
 Run `npm install` within this folder
+
+And insert your CircleCI token into the script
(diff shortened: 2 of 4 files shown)


keithamus:
Hmmm. That's quite a few files - typically we've tried to get by with just one file per plugin.

You could rewrite this with zero dependencies - using `require('http')`. All the `circleci` package does is offer a nice interface over it. Instead of:

```js
ci = new CircleCI({
  auth: "PUT THE TOKEN HERE!!!"
});
ci.getRecentBuilds({ limit: 21 }).then(...)
```

You would just do:

```js
var http = require('http');
var auth = "PUT THE TOKEN HERE!!!"
http.request('https://circleci.com/api/v1/recent-builds?limit=21&circle-token=' + auth, function (response)...).end()
```


... alternatively if you're feeling _really_ adventurous, you could write this in Python or Ruby, which (because it has synchronous APIs to deal with these things) would end up with fewer LOCs, and has the benefit of working out of the box on user's macs :smile:

keithamus:
Also - you have named the file `circleci.10s.sh` - if you're going to stick with node, I think this should be changed to `circleci.10s.js`. `.sh` should be reserved for shell scripts that have a `bash`/`sh` shebang.

mCzolko:
Let me do my best :)

tresni:
Closing for now. Fell free to submit a new PR if you do decide to revisit this

PR #197, opened 2016-01-28
Add plugin to encode current clipboard entry w/ BASE64

citoki:
This is a plugin for bitbar to take the current entry from pasteboard in OSX (a.k.a. clipboard), encode it with [BASE64](https://en.wikipedia.org/wiki/Base64) and put it back into the pasteboard.

### DIFF
--- a/System/clipboard-base64-encoder.sh
+++ b/System/clipboard-base64-encoder.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# <bitbar.title>Clipboard BASE64-Encoder</bitbar.title>
+# <bitbar.version>v0.1</bitbar.version>
+# <bitbar.author>Steffen Froehlich</bitbar.author>
+# <bitbar.author.github>citoki</bitbar.author.github>
+# <bitbar.desc>This plugin will encode acutal clipboard item with BASE64.
+# The result will be available in clipboard again. Do not use the result
+# for cryptographic purposes!!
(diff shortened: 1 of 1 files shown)


keithamus:
Happy to merge this - but IM(Humble)O you should check out [Alfred](https://www.alfredapp.com/) because I think it'd be cooler as an Alfred workflow :wink: :tophat:

citoki:
Thanks for merging! Alfred could make a lot of bitbar-plugins obsolete. And to use their clipboard integration you have to buy the 'powerpack'. Anyway, thanks for the hint, currently downloading the Alfred app.

keithamus:
@citoki I use both Alfred and BitBar (big fan of bitbar, which led me to being a collaborator on the plugins repo!). Both have their uses, and both overlap by a small amount; but to me:

 - Alfred is powerful because it lets you act on things. You can think "I want to do this" and get an alfred plugin to do it
 - BitBar is powerful because it gives you insight into things. You might not be thinking about a thing, but then bitbar comes up and says "hey, <this> is happening right now", and now you can act on it.

Action vs insight is subtle but important, IMO. As an example, the bitbar Docker plugin is very useful for people who monitor docker, I wanted that functionality but I dont need the insight, just the action - so I made https://github.com/keithamus/alfred-docker-wizard which does basically the same thing, but doesnt sit in my notifications bar :smile:

citoki:
@keithamus thanks for sharing your valuable input! Didn't really thought about Action vs Insight before, but makes totally sense.

PR #201, opened 2016-02-01
Pagerduty

frese:
Hi - I've added a PagerDuty plugin to show incidents for the past 24 hours.
Incidents are colour coded according to status and you'r able to acknowledge / resolve incidents directly in the dropdown menu.

You will need to create a config file ".pagerduty" in your home directory, with the following content.
TOKEN: xx786567x765x76
DOMAIN: my-domain.com
USERID: XXXXXX

### DIFF
--- a/PagerDuty/pagerduty.30s.rb
+++ b/PagerDuty/pagerduty.30s.rb
@@ -131,7 +131,7 @@ class PagerDuty
 
                 bash = "bash=#{File.expand_path(__FILE__)} param1=#{option} param2=#{incident['id']}" unless incident['status'].eql?("resolved")
                 puts "#{count}#{urgency} [#{incident['created_on'].split(/[TZ]/)[1]}] #{incident['incident_key']}#{urgency}|color=#{color} #{bash} terminal=false length=100"
-                puts "#{desc}|color=#{$color['normal']} length=100"
+                puts "#{desc}|color=#{$color['normal']} size=11 length=100"
                 puts "---"
             }
         end
--- a/PagerDuty/pagerduty.30s.rb
+++ b/PagerDuty/pagerduty.30s.rb
@@ -36,7 +36,7 @@ class PagerDuty
 			if ENV['BitBarDarkMode'].nil?
 				$color['normal']       = 'black'
 				$color['triggered']    = 'red'
-				$color['acknowledged'] = 'yellow'
+				$color['acknowledged'] = 'orange'
 				$color['resolved']     = 'green'
 			else
 				$color['normal']       = 'white'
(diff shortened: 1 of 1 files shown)


PagerDuty/pagerduty.30s.rb
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/ruby
+# <bitbar.title>PagerDuty</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.a uthor>Allan Frese</bitbar.author>
```
keithamus: This metadata looks a little broken. Could you fix it please?

PagerDuty/pagerduty.30s.rb
```diff
@@ -0,0 +1,158 @@
...
+# <bitbar.author.github>frese</bitbar.author.github>
+# <bitbar.desc>Shows current pagerduty alert status.</bitbar.desc>
+# <bitbar.dependencies>ruby (httparty gem)</bitbar.dependencies>
```
keithamus: It'd be awesome to see a screenshot of the plugin, linked with the `<bitbar.image>` metadata.

keithamus:
Hey @frese great work. Could you take a look at a couple of the small points above please?

PagerDuty/pagerduty.30s.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    def main
+        begin
+            # Init some defaults
+            $verbose = false
+            $id      = nil
+            $command = "GET"
+            $color   = Hash.new
+            begin
+                $config  = YAML.load(File.read("#{ENV['HOME']}/.pagerduty"))
```
keithamus: Convention with our other plugins has been to use environment variables for configuration. This is a bit of an unsolved problem right now - but I'm not sure plugins should be reading config from disk just yet. Could you please change these to environment variables instead?

frese: I would be happy to change to environment variables - but it seems very awkward  as to how you define these in various OS X releases. ! 
And right now I can't seem to figure out how to set them in El Capitan ?? I've tried .bash_profile, /etc/launchd.conf, "launchctl setenv ..." and .MacOSX/environment.plist ?
How can I set environment variables that will be available for bitbar plugins ?

keithamus: Sorry... actually I meant just standard variables at the top of the file, which consumers of the plugin just edit. Have a look at, for example, [subreedit_links.rbL6-19](https://github.com/matryer/bitbar-plugins/blob/475e51caaa0193cc0a749c768b71173fe5cc74d6/Web/Reddit/subreddit_links.rb#L6-L19) which lets you set a list of reddits to subscribe to

frese:
So, the .pagerduty configuration file is gone :-) - now the config is in the code.

I'm looking forward that your brilliant app will be able to provide plugins with configuration data :-)

frese:
I'm not able to find a suitable screenshot at the moment (with out it containing sensible info).
I'll add it as soon as I have something looking awesome :-)

frese:
@keithamus - so I've added the awesome image :-)

keithamus:
Yay! ✋

PR #207, opened 2016-02-03
Add support for IKEv2 VPN 

### LINKED ISSUE #136: re: VPN Checker
FAO: (@pldubouilh)
El Capitan now supports IKEv2 VPN, and with it a new interface. On line 12, instead of using `ifconfig | grep tun0` it should now use `ifconfig | egrep '(utun0)|(ipsec0)'` or `ifconfig | grep 'utun0\|ipsec0'` so it can accurately detect both types of VPN connections.

This will also need to be changed for line 14 at `ifconfig utun0 2> /dev/null | grep inet | cut -d' ' -f 2` but I couldn't quite figure out how to do that.

pldubouilh:
Should fix #136


matryer:
LGTM

PR #208, opened 2016-02-03
Add network location plugin

ffittschen:
This plugin shows the current network location and makes it possible to change to a different one.

Screenshot


matryer:
Beautiful

PR #209, opened 2016-02-03
BAJ update, adding BGO

astrostl:


keithamus:
Hey @astrostl thanks for the PR! But I have a question: If its the same script with the exception of a different domain, why not either call the baj script from bgo - or have bgo has a variable in the baj script?

astrostl:
In addition to the names they have separate URLs, forms, and REs:

```
-url="http://www.boiteajeux.net/gestion.php"
-data="pAction=login&password=${baj_password}&username=${baj_login}"
+url="http://www.boardgaming-online.com/"
+data="&identifiant=${bgo_login}&mot_de_passe=${bgo_pass}"
```

```
-extract() { sed -n '/<title>/{s/.*\[\([0-9]*\)\].*/\1/p;q;}'; }
+extract() { sed -n '/^nbp=/{s/nbp=\([0-9]*\);/\1/p;q;}'; }
```

While they're similar in that they both curl-and-process, I think it makes sense as two different scripts.

keithamus:
Ok cool :smile: I shall merge.

PR #211, opened 2016-02-05
add ip-flag plugin

MatzFan:


keithamus:
Hey @MatzFan thanks for the PR!

Seems like this is quite a few requests than happen every 2 seconds, plus http://data.okfn.org/data/core/country-list/r/data.csv seems to be returning an error page for me. Perhaps you could have a simple enumeration of ipinfo country codes to emoji, and change the file to run every 30s instead?

keithamus:
Also if you simply curl `curl ipinfo.io/country` you'll get the 2 letter code, rather than having to pipe through grep+sed.

keithamus:
@MatzFan I had a little play. Unicode country flags are actually really cool.

Country flags are composed of 2 utf-16 characters called "regional indicator symbols", they start from A (http://www.fileformat.info/info/unicode/char/1f1e6/index.htm) and go to Z (http://www.fileformat.info/info/unicode/char/1f1ff/index.htm). The country for for US for example, is comprised of `Unicode Regional Indicator Symbol U` + `Unicode Regional Indicator Symbol S`. To break that down into UTF-8 bytes, that is `F0 9F 87 BA F0 9F 87 B8`. Similarly, GB (England's country code) is `F0 9F 87 AC F0 9F 87 A7`

So we can use a little bit of number wrangling to _generate_ the country code, without using a hash map or anything! We just need to get the ISO-2 country code (which, happily, ipinfo provides) and then take each letter's unicode symbol and add `0x65` (capital letters start at `0x41`, our regional codes start at (a bunch of bytes plus) `0xA2`). 

Here's a little python script I've chucked together which does the same thing:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

# <bitbar.title>External IP (think VPN) country flag emoji</bitbar.title>
# <bitbar.version>v1.5 beta 1</bitbar.version>
# <bitbar.author>Bruce Steedman</bitbar.author>
# <bitbar.author.github>MatzFan</bitbar.author.github>
# <bitbar.desc>Displays country flag emoji - useful if you switch VPN's & need to know remember you 'are' ;-)</bitbar.desc>
#<bitbar.dependencies>OS X 10.11, Sadly Tor breaks it due to rotten CloudFlare Captchas..</bitbar.dependencies>

import json
import urllib2

code = urllib2.urlopen("http://ipinfo.io/country").read()
firstLetterChr = chr(ord(code[0]) + 0x65);
secondLetterChr = chr(ord(code[1]) + 0x65);
print '\xF0' + '\x9F' + '\x87' + firstLetterChr + '\xF0' + '\x9F' + '\x87' + secondLetterChr
```

What do you think? Happy to update the PR to feature this code instead?

manojlds:
@keithamus wow, that's cool :+1:

MatzFan:
Of course @keithamus, this is a big improvement. Had in fact realised you can generate flags from UTF-8 after I sent the PR! Will update it shortly.

keithamus:
#212

MatzFan:
Replaced with [#212](https://github.com/matryer/bitbar-plugins/pull/212) Ruby script

MatzFan:
Thanks Keith - I learnt something from your great script, but Ruby is more my thing ;)

keithamus:
That's totally fine, to each their own :smile:

PR #213, opened 2016-02-05
2m interval to ensure within free api call limit

MatzFan:


keithamus:
Ah good catch, missed that!

PR #215, opened 2016-02-09
Plugin for Sentry

Nevon:
https://getsentry.com

A crash reporting service.

screen shot 2016-02-09 at 15 47 16


keithamus:
:+1: good work

PR #217, opened 2016-02-09
last.fm now playing

estiens:
Ruby script for grabbing now playing/last played information from a Last fm account

image


keithamus:
💯👍

PR #218, opened 2016-02-09
Added Bitfinex BTCUSD Price

mileschet:
Getting BTCUSD from Bitfinex public rest api, more info at: http://docs.bitfinex.com/#rest


keithamus:
Good work :+1:

PR #219, opened 2016-02-10
add plugin to detect if firewall is enabled

eddietejeda:


System/firewall.10m.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# <bitbar.title>Firewall</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Eddie A Tejeda</bitbar.author>
+# <bitbar.author.github>eddietejeda</bitbar.author.github>
+# <bitbar.desc>This plugin displays the status of the firewall.</bitbar.desc>
+# <bitbar.dependencies></bitbar.dependencies>
```
keithamus: This looks great, it'd be even greater if you could add the `<bitbar.image>` metadata, with an appropriate screenshot :wink:

eddietejeda: Done!

PR #221, opened 2016-02-10
Use token and fix issue with tailing json

oliviernt:
- Removes usage of password in favour of token
- Fixes issue with tailing the end of the file as JSON keys are unsorted


keithamus:
LGTM :+1:

PR #223, opened 2016-02-10
Plugin to display number of connected players on a Minecraft server

pallih:
if any are connected, then display their names in the drop-down.


tresni:
Hi, we are linting all PR's with an appropriate language linter.  In the case of python we use pyflakes.

Might make sense to set `PORT = 25565` since that's the default port with a comment to change if people are running on a custom port ;-)

```
Projects/bitbar-plugins ●» ./.test.py Games/minecraftplayers.1m.py                                                                                                                                                                                                                                                                                              « [master●]
/Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/Games/minecraftplayers.1m.py:23: undefined name 'PORT'
ERR! /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/Games/minecraftplayers.1m.py failed linting, see above errors

ERR! failed with 1 errors

Projects/bitbar-plugins ●»
```

pallih:
Good point! Should be fixed.

PR #224, opened 2016-02-10
Create next_train.1m.sh

danburke:
Added a script to retrieve the next set of trains from Irish Rail for a particular station. 
Depends on jq - https://stedolan.github.io/jq/
Preview Image

### DIFF
--- a/Web/IrishRail/next_train.1m.sh
+++ b/Web/IrishRail/next_train.1m.sh
@@ -13,6 +13,6 @@ STATIONID='PERSE'
 # Replace the DIRECTION with the appropriate DIRECTION
 DIRECTION='Southbound'
 json=$(curl -s https://dublin-trains.herokuapp.com/api/stations/$STATIONID/trains |  /usr/local/bin/jq -r '.["'$DIRECTION'"]')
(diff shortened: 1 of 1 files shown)


Web/IrishRail/next_train.1m.sh
```diff
@@ -0,0 +1,18 @@
...
+# Depends on jq - https://stedolan.github.io/jq/
+
+# Replace the STATIONID with the appropriate ID
+STATIONID='PERSE'
+# Replace the DIRECTION with the appropriate DIRECTION
+DIRECTION='Southbound'
+json=$(curl -s https://dublin-trains.herokuapp.com/api/stations/$STATIONID/trains |  /usr/local/bin/jq -r '.["'$DIRECTION'"]')
+echo $json | /usr/local/bin/jq -r 'map(select(.scheduled | contains ("--") | not) )| .[0].last_update'
+echo '---'
+echo $json | /usr/local/bin/jq -r 'map(select(.scheduled | contains ("--") | not) )| .[] | "\(.destination): \(.last_update)"'
```
keithamus: The two echo lines above, where you `echo $json` - these should be `echo "$json"` otherwise bash can do some silly things.

Issue #198, opened 2016-01-29
re: Weather


MichaelPak:
FAO: (@seripap)

    michaelpak@macbook: /Users/michaelpak/Documents/BitBarPlugins/weather.15m.py
    Could not get weather

MichaelPak:
I have already got `api_key`, but it still doesn't work

MichaelPak:
I have found problem. It's in `get_wx()` in `weather_data['visibility']` and `weather_data['next_hour']`, cause `json` response from `forecast` hasn't got `['currently']['visibility']` and `['minutely']['summary']`

seripap:
What location are you grabbing weather for? The plugin checks weather for all those nodes- if one doesn't exist it'll catch the exception and hang. You can modify it by deleting some of the nodes that may not work for specific location, or add better error-handling for each of the nodes

MichaelPak:
That's response from `ipinfo.io`:

    {
      "ip": "212.5.****.****",
      "hostname": "firewall.****.ru",
      "city": "",
      "region": "",
      "country": "RU",
      "loc": "55.7500,37.6166",
      "org": "****"
    }

I'm from Russia/Moscow

chakler:
I live in Manila and have the same problem as well. The JSON I got from forcast.io doesn’t provide any `minutely` information.

It’d probably be better if the less important information are conditional based on availability. :smiley:

seripap:
I'll fix this up to be a bit more conditional for my international friends :+1:

MichaelPak:
Thanks a lot!

chakler:
Tried https://github.com/matryer/bitbar-plugins/pull/226. Works like a charm! Thanks!

MichaelPak:
good work amigo!

PR #226, opened 2016-02-10
Weather v2.0.0

seripap:
* Solution for #198 
* Better support for international location
* Better error handling/messaging
* Support for URL opening in bitbar v1.4+


tiagofassoni:
Hey! I used your PR, and there is a problem with it if the city name contains accents (like 'São Paulo'). The problem is that in Python 2 the function str() expects ASCII. I changed it to unicode() and it worked flawlessly. I'd love to fix it in the PR, but I have absolutely no idea on how to do a PR inside a PR (PR-ception, maybe?)

seripap:
@tiagofassoni Just updated to encode everything into utf-8- does that fix it? I'm running python 2.7 and everything seems to work

tiagofassoni:
Yes, it works, thanks!

PR #229, opened 2016-02-10
Script to show random quotes created

JanGross:
This script grabs a list of quotes and shows a random one to the user on each refresh.


JanGross:
Image URL corrected in 749b934

PR #232, opened 2016-02-11
If area does not have hourly information, plugin does not fail

tiagofassoni:
My area does not have hourly information. That case isn't handled by the script, which just fails. I added a conditional that deals with that case.


seripap:
Please do not merge this, #226 enhances this plugin.

tiagofassoni:
Okay, I'm cancelling this pull request. By the way, #226 is really nice!

PR #235, opened 2016-02-11
Homebrew Cask Updater

Srdgh:
New Plugin (Homebrew Cask Updater)
+ slight tweak to caffeinate plugin 
+ slight tweak to Todo Colour


Dev/Homebrew/homebrewcask.1d.sh
```diff
@@ -0,0 +1,25 @@
...
+brewcall=$(/usr/local/bin/brew cask ls -1 | sed 's/(!)//g' | xargs /usr/local/bin/brew cask info | grep -A 1 'Not installed' | sed -e 's_Not installed__g' -e 's_https://github\.com/caskroom/homebrew-cask/blob/master/Casks/__g' -e 's_\.rb__g')
+
+brewnum=$(for line in $brewcall; do echo $line | grep "[a-z]" ; done | wc -w)
+
+echo "${brewnum}"
+echo "---"
+echo $brewnum "casks to update"
+if [[ "${brewnum}" -ne "0" ]]
+then 
+echo "Click item to install new version:"
```
matryer: Instead of this, how about you just prefix the other lines so it looks like this:

```
2 casks to update
update rstudio
update unetbootin
```

Slightly smaller menu :)

Srdgh:
Good point! -- done.

matryer:
I love this plugin. What does it do when there are no plugins? I think it should output nothing, which will make BitBar hide the menubar altogether.

Also, is the 3 enough in the Menu bar, or should it say `3 updates`?

Sorry to be a pain, but I think this could be how we advertise people when a new BitBar release is available... and maybe even pre-package this plugin.

Srdgh:
Aw thanks! I love BitBar!
- I've modified the menubar counter from "3" to "3 updates". (Personally, this is more wordy than I need, but I agree it may be more approachable for first-time user.)
- Plugin now outputs nothing when there are no casks to update.
- Added some further Brew commands, commented out, so the user can add these commands to the drop-down by un-commenting them.

Happy to tweak and modify as you suggest. Would be right proud if you ended up using this for the new release.

Srdgh:
A leaner alternative to outputting "3 updates" into the menubar might be to use the beer mug emoji. Two possibilities:
(1) The beer mug emoji plus number of updates: `🍺3`
(2) The beer mug emoji for one update (`🍺`) and clinking beer mugs emoji (`🍻`) if there are more than one update.

My preference is for solution (2). It is even less cumbersome than (1), and I think what people will want to get from this plugin is /that/ they have cask-updates, not /how many/ cask-updates they have. They can find out how many they have from the drop-down.

I'll commit with this solution (2), but if you disagree I'm happy to revert to the "3 updates" or to solution (1).

vogonistic:
I love (2) with :beer: and :beers:!

tresni:
You aren't executable.  Need to `chmod +x` your script.  That's why it's failing the tests :)

tresni:
Also, I'd appreciate it if you separated the plugins out into different Pull Requests and squash commits at this point.  Makes it easier to merge things that are passing ;-)

Srdgh:
Thanks for the advice -- now done. Let me know if you need me to do anything else.

tresni:
Squashed your commits and merged as 2da40df.

PR #241, opened 2016-02-13
deathTimer added

twolfe2:
I created this plugin for myself so that I could be reminded that death is always near. This plugin helps to keep me motivated and I hope it will help keep others motivated too. 

This plugin will allow a user to input the estimated year they will die and then it will provide them with a countdown to that year.


Lifestyle/deathTimer.py
```diff
@@ -0,0 +1,46 @@
...
+# <bitbar.desc>Counts down how approximately how many days you have left to live</bitbar.desc>
+# <bitbar.image>http://imgur.com/2KjZaFb</bitbar.image>
+# <bitbar.dependencies>python</bitbar.dependencies>
+
+
+
+from datetime import datetime
+from datetime import timedelta
+from datetime import datetime
+from datetime import timedelta
```
keithamus: Looks like you're importing things twice here.

twolfe2: Fixed, thanks.

keithamus:
Oh, also I just noticed the file is called `deathTimer.py` - but should probably be `deathTimer.1s.py` or something?

twolfe2:
Since right now it will be displaying the approximate number of days remaining I will have it update every 5 minutes. In a future update I might have it display days, hours, minutes, and seconds remaining.

keithamus:
What a wonderfully somber plugin btw :smile:.

PR #247, opened 2016-02-15
Plugin to show colour coded trackpad battery % 

milch:


keithamus:
I think this might be our first swift plugin!

PR #249, opened 2016-02-17
Add countdown timer plugin

kizzx2:


matryer:
Have you seen this? https://getbitbar.com/plugins/Time/countdown.1s.py

kizzx2:
@matryer 

Yes. That one is for counting down to a specific date time which you must change by editing the source file manually.

This one is for firing timer for "one off" tasks, examples here https://github.com/kizzx2/bitbar-countdown-timer. The main motivation is to do time-boxed tasks.

A more similar one is actually https://getbitbar.com/plugins/Time/pomodoro.1s.sh but that is fixed at 25 minutes time boxes.

PR #257, opened 2016-02-19
add java_version plugin

ssanj:
This plugin displays the current version of Java installed on your
system. If it is the newest version it will be displayed in green
on the task bar. If there are newer versions, your current Java
version will be displayed in orange and the context menu will have
a list of newer versions which you can click on.

Clicking on a newer version takes you to a download page which will
let you download and install the latest Java version.


ssanj:
Hi,

I noticed this plugin does not show up under the [plugins page](https://getbitbar.com/plugins/Dev). It looks like I've added all the necessary meta tags. Do I need to do something else for this to happen?

PR #258, opened 2016-02-19
Timestamp to date

mpneuried:
I added a script to convert unix timestamps to a date.

Open Features:
- detect and switch between seconds and miliseconds

### DIFF
--- a/Time/ts-convert.sh
+++ b/Time/ts-convert.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# <bitbar.title>Clipboard Timestamp converter</bitbar.title>
+# <bitbar.version>v0.1</bitbar.version>
+# <bitbar.author>M. Peter</bitbar.author>
+# <bitbar.author.github>mpneuried</bitbar.author.github>
+# <bitbar.desc>This plugin converts timestamps from the clipboard to a date.
+# It's inspired by the original clipboard-base64-encoder plugin</bitbar.desc>
+# <bitbar.image></birtbar.image>
+# <bitbar.dependencies></bitbar.dependencies>
+
+# Hack for language not being set properly and unicode support
+export LANG="${LANG:-en_US.UTF-8}"
+
+PREFIX=""
+APPENDIX=""
+
+
+# user clicks on 'encodepaste'
+if [[ "$1" = "encodepaste" ]]; then
+  CONVERTED=$(echo -n "${PREFIX}$(pbpaste)${APPENDIX}" | xargs -I {} date -r {} +'%Y-%m-%d %H:%M:%S')
+  $(echo -n "$CONVERTED" | pbcopy)
+  osascript -e "display notification \"$CONVERTED\" with title \"DATE:\"" &> /dev/null
+  exit
+fi
+
+# user clicks on 'encoding'
+if [[ "$1" = "encode" ]]; then
+  CONVERTED=$(echo -n "${PREFIX}$(pbpaste)${APPENDIX}" | xargs -I {} date -r {} +'%Y-%m-%d %H:%M:%S')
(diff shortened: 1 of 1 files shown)


Time/ts-convert.sh
```diff
@@ -0,0 +1,48 @@
...
+export LANG="${LANG:-en_US.UTF-8}"
+
+PREFIX=""
+APPENDIX=""
+
+
+# user clicks on 'encodepaste'
+if [[ "$1" = "encodepaste" ]]; then
+  CONVERTED=$(echo -n "${PREFIX}$(pbpaste)${APPENDIX}" | xargs -I {} date -r {} +'%Y-%m-%d %H:%M:%S')
+  $(echo -n "$CONVERTED" | pbcopy)
```
keithamus: This line should not include `$()` because that attempts to execute any output.

Time/ts-convert.sh
```diff
@@ -0,0 +1,48 @@
...
+  CONVERTED=$(echo -n "${PREFIX}$(pbpaste)${APPENDIX}" | xargs -I {} date -r {} +'%Y-%m-%d %H:%M:%S')
+  osascript -e "display notification \"$CONVERTED\" with title \"DATE:\"" &> /dev/null
+  exit
+fi
+
+# user clicks on 'now'
+if [[ "$1" = "now" ]]; then
+  NOW=$(date +%s)
+
+  $(echo -n "$NOW" | pbcopy)
```
keithamus: This line should not include `$()` because that attempts to execute any output.

mpneuried:
i changed it to:
`CONVERTED=$(date -r `pbpaste` +'%Y-%m-%d %H:%M:%S')`

Hopefully this is a better solution ;-)

Time/ts-convert.sh
```diff
@@ -0,0 +1,43 @@
...
+# <bitbar.image>http://i.imgur.com/dXzJNOS.jpg?1</birtbar.image>
+# <bitbar.dependencies></bitbar.dependencies>
+
+# Hack for language not being set properly and unicode support
+export LANG="${LANG:-en_US.UTF-8}"
+
+# user clicks on 'encodepaste'
+if [[ "$1" = "encodepaste" ]]; then
+  CONVERTED=$(date -r `pbpaste` +'%Y-%m-%d %H:%M:%S')
+  $(echo -n "$CONVERTED" | pbcopy)
```
keithamus: This still needs to loose the $(). So

```sh
$(echo -n "$CONVERTED" | pbcopy)
```

Needs to become
```sh
echo -n "$CONVERTED" | pbcopy
```

Time/ts-convert.sh
```diff
@@ -0,0 +1,43 @@
...
+if [[ "$1" = "encode" ]]; then
+  CONVERTED=$(date -r `pbpaste` +'%Y-%m-%d %H:%M:%S')
+  osascript -e "display notification \"$CONVERTED\" with title \"DATE:\"" &> /dev/null
+  exit
+fi
+
+# user clicks on 'now'
+if [[ "$1" = "now" ]]; then
+  NOW=$(date +%s)
+  $(echo -n "$NOW" | pbcopy)
```
keithamus: Same here.

mpneuried:
@keithamus Thanks for you help.
I'm very new to bash scripting and thankful for every hint ;-)

But one question... I used [this](https://github.com/matryer/bitbar-plugins/blob/master/System/clipboard-base64-encoder.sh#L22) as inspiration.
```sh
$(echo -n "$(echo -n "${PREFIX}$(pbpaste)${APPENDIX}" | base64)" | pbcopy)
```
Because this looks quite similar i guess it has the same issues?

keithamus:
Yeah, I probably shouldn't have merged that like it is - that will also have the same problems.

PR #270, opened 2016-02-23
Rewrote the uptime plugin for easier customization

chamini2:
Added *flags* `PLURAL` for plural or singular output, and `VERBOSE` for including some values even when they are 0.
The variables `DS`, `HS`, `MS`, `SS` and `SEP` can be modified to change the output to own preference.

**** 

As an example, the user may prefer:

    PLURAL = 0

    SEP = " : "

    DS = "d"
    HS = "h"
    MS = "m"
    SS = "s"

To get output like:
> [ ↑ 2d : 10h : 5m ]

****

Or in another language, in spanish:

    PLURAL = 1

    SEP = ", "

    DS = " día"
    HS = " hora"
    MS = " minuto"
    SS = " segundo"

To get output like:
> [ ↑ 2 días, 10 horas, 5 minutos ]

### DIFF
--- a/System/uptime.1m.sh
+++ b/System/uptime.1m.sh
@@ -9,20 +9,6 @@
 
 INFO=`uptime`
 echo $INFO | awk -F'[ ,:\t\n]+' '
-    function include(VAL, UNIT, SUFFIX, PLURAL) {
-        VAL = int(VAL)
-
-        if (PLURAL && VAL > 1) {
-            UNIT = UNIT"s"
-        }
-
-        if (VAL > 0) {
-            return (VAL UNIT SUFFIX)
-        } else {
-            return ""
-        }
-    }
-
     {
         PLURAL = 1
 
@@ -71,6 +57,19 @@ echo $INFO | awk -F'[ ,:\t\n]+' '
         MSG = substr(MSG, 0, length(MSG) - length(SEP))
 
         print "[", MSG, "] | size=12"
+    }
+    function include(VAL, UNIT, SUFFIX, PLURAL) {
+        VAL = int(VAL)
+
+        if (PLURAL && VAL > 1) {
+            UNIT = UNIT"s"
+        }
+
+        if (VAL > 0) {
+            return (VAL UNIT SUFFIX)
+        } else {
+            return ""
+        }
     }'
 
 echo "---"
--- a/System/uptime.1m.sh
+++ b/System/uptime.1m.sh
@@ -24,6 +24,17 @@ echo $INFO | awk -F'[ ,:\t\n]+' '
     }
 
     {
+        PLURAL = 1
+
+        SEP = ", "
+
+        DS = " day"
+        HS = " hr"
+        MS = " min"
+        SS = " sec"
+
+        ################################
+
         D = H = M = S = 0
 
         if (substr($5,0,1) == "d") {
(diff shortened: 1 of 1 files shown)


chamini2:
As an example, the user may prefer:

    PLURAL = 0

    SEP = " : "

    DS = "d"
    HS = "h"
    MS = "m"
    SS = "s"

To get output like:
> [ ↑ 2d : 10h : 5m ]

tresni:
Please address the following linting reports from shellcheck

```
Projects/bitbar-plugins ●» ./.test.py System/uptime.1m.sh                                                                                                                                                                                                                                                                                                       « [master●]

In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 10:
INFO=`uptime`


In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 11:
echo $INFO | awk -F'[ ,:\t\n]+' '


In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 52:
        MSG = "↑ " include(D, DS, SEP, PLURAL) \
                                                ^-- SC1004: You don't break lines with \ in single quotes, it results in literal backslash-linefeed.


In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 53:
                   include(H, HS, SEP, PLURAL, (D > 0 && VERBOSE)) \
                                                                    ^-- SC1004: You don't break lines with \ in single quotes, it results in literal backslash-linefeed.


In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 54:
                   include(M, MS, SEP, PLURAL, (D > 0 && VERBOSE)) \
                                                                    ^-- SC1004: You don't break lines with \ in single quotes, it results in literal backslash-linefeed.


In /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh line 78:
echo $INFO | tr "," "\n" | tail -n 2

ERR! /Volumes/Home/bhartvigsen/Documents/Projects/bitbar-plugins/System/uptime.1m.sh failed linting, see above errors

ERR! failed with 1 errors

Projects/bitbar-plugins ●»
```

chamini2:
@tresni, fixed the linting reports. Let me know if it needs anything else done.

v8media:
Would there be an easy way to modify the script to only show an uptime once it gets past a certain amount of days? It would be great to have another variable for this.

Is BitBar able to show nothing, or maybe only use 1 pixel of width in this type of scenario?

gingerbeardman:
You can print a space.

I do this with the GitHub BitBar plugin. It looks like this:

Screen shot 2019-08-30 at 12 31 08

gingerbeardman:
Replace:
https://github.com/matryer/bitbar-plugins/blob/ec3ac184198936271503018330314fa38ad63a1a/System/uptime.1m.sh#L60

With (for only showing if uptime is more than 4 days):
```sh
if (D > 4) {
    print "[", MSG, "] | size=12"
} else {
    print " "
}
```

v8media:
Thanks! That's great! I think I may put something like this on all of my
clients machines so that they get an indicator to restart when their
machines have been up for a long time.

gingerbeardman:
I'm not sure I agree not that, but you know your clients better than I do!

PR #286, opened 2016-02-28
Added new plugin to check status of Linode(s)

misfitius:
A bitbar plug-in to show the status of your Linode(s). Requires linode-cli to be installed (https://www.linode.com/docs/platform/linode-cli).

### DIFF
--- a/Linode.5m.php
+++ b/Linode.5m.php
@@ -1,0 +1,23 @@
+#!/usr/bin/php
+
+<?php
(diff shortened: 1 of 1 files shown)


tresni:
2 issues:

1. The script is not executable.  Please `chmod +x` the script.
2. It is just in the root Plugin directory, please put it under an appropriate subdirectory (may need to create something like VPS/Linode/Linode.5m.php)

PR #289, opened 2016-02-29
Adding new bitbar version check plugin

oliviernt:
Adding new version plugin for BitBar which checks current installed version against latest from GitHub

### DIFF
--- a/BitBar/version.12h.rb
+++ b/BitBar/version.12h.rb
@@ -17,12 +17,16 @@ require 'net/http'
 require 'json'
 require 'nokogiri'
 
+# if you're seeing errors saying you've reached the API request limit you will need to
 # create a new application at https://github.com/settings/developers and add client_id and client_secret here:
-GITHUB_CLIENT_ID=""
-GITHUB_CLIENT_SECRET=""
+# GITHUB_CLIENT_ID=""
+# GITHUB_CLIENT_SECRET=""
+# 
+# then add the following query params to the github URL
+# ?client_id=#{GITHUB_CLIENT_ID}&client_secret=#{GITHUB_CLIENT_SECRET}
 
 def get_json
-  url = "https://api.github.com/repos/matryer/bitbar/releases/latest?client_id=#{GITHUB_CLIENT_ID}&client_secret=#{GITHUB_CLIENT_SECRET}"
+  url = "https://api.github.com/repos/matryer/bitbar/releases/latest"
   json_result = JSON.parse(Net::HTTP.get(URI(url)))
   json_result
 end
--- a/BitBar/version.12h.rb
+++ b/BitBar/version.12h.rb
@@ -21,7 +21,7 @@ require 'nokogiri'
 # create a new application at https://github.com/settings/developers and add client_id and client_secret here:
 # GITHUB_CLIENT_ID=""
 # GITHUB_CLIENT_SECRET=""
-# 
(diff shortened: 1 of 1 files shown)


BitBar/version.12h.rb
```diff
@@ -0,0 +1,63 @@
...
+# BitBar version plugin
+# by Olivier Tille (@oliviernt)
+#
+# Checks the current BitBar version against the latest from GitHub
+
+require 'net/http'
+require 'json'
+require 'nokogiri'
+
+# create a new application at https://github.com/settings/developers and add client_id and client_secret here:
```
iosdeveloper: It seems to be a public endpoint. I don't think you need these.

oliviernt: You're absolutely right. I had that in there for development as I would hit the max request limit but that shouldn't be a problem when simply using the plugin.

keithamus:
Hey @oliviernt! You have unwittingly become the first of our contributors to hit our test suite! There are some minor issues that need addressing before we can merge this PR. I will add them as notes to the changeset.

BitBar/version.12h.rb
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env ruby
```
keithamus: This file is not executable. Please `chmod +x` and commit the permissions (`git add BitBar/version.12h.rb`)

BitBar/version.12h.rb
```diff
@@ -0,0 +1,67 @@
...
+def get_json
+  url = "https://api.github.com/repos/matryer/bitbar/releases/latest"
+  json_result = JSON.parse(Net::HTTP.get(URI(url)))
+  json_result
+end
+
+def get_xml
+  bitbar_path = `osascript -e 'tell application "System Events" to POSIX path of (file of process "BitBar" as alias)'`.chomp
+  bitbar_path += "/Contents/Info.plist"
+  doc = File.open(bitbar_path) { |f| Nokogiri::XML(f) }
```
keithamus: `doc` is not needed, it is a "useless assignment.

See:
```rb
https://travis-ci.org/matryer/bitbar-plugins/builds/116564908#L300-L302
BitBar/version.12h.rb:37:3: W: Useless assignment to variable - doc.
  doc = File.open(bitbar_path) { |f| Nokogiri::XML(f) }
  ^^^
```

BitBar/version.12h.rb
```diff
@@ -0,0 +1,67 @@
...
+  outdated = Gem::Version.new(current_version) < Gem::Version.new(latest_version.sub!("v", ""))
+  color = outdated ? "red" : "green"
+
+  puts current_version + " | color=" + color
+
+  if outdated
+    puts "---"
+    puts "Download latest (#{latest_version}) | href=" + json_val["assets"][0]["browser_download_url"]
+  end
+rescue => e
```
keithamus: `e` is a "useless assignment". You can change it to `_` and it will not give back a lint error.

See:
https://travis-ci.org/matryer/bitbar-plugins/builds/116564908#L303-L305
```rb
BitBar/version.12h.rb:63:11: W: Useless assignment to variable - e.
rescue => e
```

BitBar/version.12h.rb
tresni: Fixed in 1cd0cf4

PR #290, opened 2016-02-29
Tutorial plugin for image parameters

ibutra:
As discussed in https://github.com/matryer/bitbar/pull/261 this is the tutorial for the `templateImage` and `image` parameters


Tutorial/images.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+#
+# Include BitBar metadata like this at the top of the file
```
iosdeveloper: This isn't the scope of this tutorial and should be removed.

Tutorial/images.sh
```diff
@@ -0,0 +1,28 @@
...
+# (commented out, of course):
+#
```
iosdeveloper: This isn't the scope of this tutorial and should be removed.

Tutorial/images.sh
```diff
@@ -0,0 +1,28 @@
...
+# <bitbar.title>Show templateImages and imagesd</bitbar.title>
```
iosdeveloper: The last character shouldn't be there.

keithamus:
Shall we merge this when `templateImages` are actually released?

iosdeveloper:
@keithamus [https://github.com/matryer/bitbar/releases/tag/v1.7.0](v1.7.0) was released with image parameters added.

iosdeveloper:
@matryer Can we please get this merged?

tresni:
Looks good to me. Merging.

PR #291, opened 2016-03-02
Get rid of Spotify's redundant info from track title

jostyee:
e.g.:
" - 2012 Remastered Version"
" - Single Version"


tresni:
I'm going to leave this one open to see what people who actually use spotify think.  If you like it, please add a reaction emoji.

dkhamsing:
Don't come across this much but seems reasonable

PR #302, opened 2016-03-08
Added netinfo plugin. 

q1x:
It provides info on internal and external IP addresses (v4 and v6) for each adapter, Whois information and Speedtest.net results.


q1x:
I've updated the plugin and will submit a new pull request

PR #326, opened 2016-03-17
Speeding up external IP discovery by using dig and opendns

brunomvsouza:
A DNS lookups are faster than http requests.

### DIFF
--- a/Network/external-ip.1h.sh
+++ b/Network/external-ip.1h.sh
@@ -7,7 +7,7 @@
 #
 # Gets the current external IP address.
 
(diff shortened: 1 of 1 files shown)


tresni:
Full disclosure: I work for OpenDNS :)  I :heart: this but I may be biased.

I don't think we should require Travis-CI complaince since it's the original code that's failing and not you're actual changes.  But I'd like @matryer or @keithamus to :+1: this first.

keithamus:
Seems fine by me.

PR #327, opened 2016-03-17
Added exception handling for ip-flag

chamini2:
This is for when there's no internet connection or stuff like that.

We could use 🚩 or 🏁, since they're flags that do not belong to a country. I prefered 🚩 because red is used for warnings / errors.


tresni:
Please review the Travis-CI build and correct the linter issues detected by Rubocop.  Thanks!

PR #332, opened 2016-03-19
add Average Review Times Plugin

mfks17:
This is a plugin to show average review times iOS & Mac App Store.
See also [my repository](https://github.com/mfks17/bitbar-plugin-AppStore).


iosdeveloper:
Nice! Please use `raw` instead of `blob` for the image URL.

mfks17:
I fixed image Metadata to use the raw file.

iosdeveloper:
👍

tresni:
Should this be under Dev instead of Web? Seems more geared towards development then general web stuff. @matryer opinion?

tresni:
Merged but moved to Dev/AppStore.

mfks17:
Thanks!  🙏

PR #335, opened 2016-03-20
Add plugin emacs-package-updates

masutaka:
This plugin gives the below information.


tresni:
Squashed and merged as 7486d81.  I removed the abouturl as it shouldn't reference the bitbar-plugins github or website.  If you have your own website that you want to point it at, that would be an appropriate use.

iosdeveloper:
@tresni The bitbar-plugins GitHub is used as about URL all over the place. ¯\_(ツ)_/¯

masutaka:
@tresni Thanks a lot!

tresni:
Maybe I'm just blind then.  Recirculated referencing seems weird though.

PR #336, opened 2016-03-20
Bit improve CI performance

masutaka:
I think gem document is no need in CI.


tresni:
This has really speed up build time. Thank you for this!

masutaka:
Welcome! I'm glad to hear it!

PR #344, opened 2016-03-21
Support dlite

eminetto:
I'm using Dlite (https://github.com/nlf/dlite) instead of docker-machine and i just added support to this feature


keithamus:
@eminetto there are some linting errors it seems.

If you switch the shebang to `#!/usr/bin/env bash` then it should fix those errors, and we can merge :smile:

eminetto:
Thanks @keithamus. I think it's ok now :)

PR #353, opened 2016-03-26
compatibility fix

Srdgh:
using "bash=... param1=..." notation rather than putting arguments into "bash=...". For compatibility with bitbar v1.9


tresni:
Just an FYI that the behavior in 1.9 is a regression that is fixed in the latest commit to BitBar. I'd expect a new build shortly.

Issue #180, opened 2016-01-25
Head to head: The best Pomodoro plugin


matryer:
We need to select between these two excellent Pomodoro plugins:

  * https://github.com/matryer/bitbar-plugins/pull/177
  * https://github.com/matryer/bitbar-plugins/pull/178

We only want one in the repo - but which one? We'll let the community decide.

keithamus:
#178:
 - + Emoji
 - + Notifications
 - + Configurable
 - - Writes multiple files
 - - Needs shellcheck changes

#177:
 - + Code more readable
 - + Shellchecked (now)
 - + Original

manojlds:
#178 :+1:

gorangajic:
I have searched plugins for pomodoro but forget to look up in pull requests :smile: what are the changes that 2 people are working on the same thing in the same time ? 

Imo we should merge both plugins because functionality is a bit different #178 have two timers(work and break) that switch automatically, and #177 have just one timer that need to be restarted manually 

also #178 play a sound when timer reach 00:00

> Writes multiple files

I can move that in one file but wanted to keep things more simpler
 
btw #177 have bug, config files should be moved from plugin directory because bitbar treat that file like a new plugin so there is new warning icon ⚠️ 

no hard feelings, whatever you guys decide :smile:

matryer:
What's funny is - a friend of mine has started work on a Pomodoro plugin too - I've sent him a link to this. :)

matryer:
Shall we spellcheck it and go with #178 then?

keithamus:
Sounds like an idea.

sagarjauhari:
Ahh, found this just when I was about to open a PR for a Pomodoro plugin I wrote too! https://github.com/sagarjauhari/pomodoro.rb/blob/master/pomodoro_bitbar.rb :smile:

PR #362, opened 2016-03-30
Adding a single-click python plugin to run caffeinate

chrisidefix:
Even though there already are two caffeinate plugins - I am submitting this pull request to add an option that is using the single-click-action feature. It is a very simplistic approach that allows you with a single click to either keep your computer awake or allow it to go to sleep.

### DIFF
--- a/System/stayawake.1h.py
+++ b/System/stayawake.1h.py
@@ -6,7 +6,7 @@
 # <bitbar.author>Christoph Russ</bitbar.author>
 # <bitbar.author.github>christophruss</bitbar.author.github>
 # <bitbar.desc>Keep your computer awake with a single click.</bitbar.desc>
-# <bitbar.image>https://imgur.com/pkqXw8x</bitbar.image>
+# <bitbar.image>http://i.imgur.com/pkqXw8x.jpg</bitbar.image>
 # <bitbar.dependencies>python</bitbar.dependencies>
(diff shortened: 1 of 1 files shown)


keithamus:
Thanks @chrisidefix, but why not modify one of the existing ones? I'm sure @citoki or @dkhamsing would happily discuss merging these plugins. While I like your code, and its a great idea to have it single-click - I think we should avoid adding too many of the same plugin to this repo.

chrisidefix:
I wasn't sure, because there are already two very similar `caffeinate` plugins by the same author and I assumed there was a reason for that. I didn't want to mess with those plugins (e.g. change the icons and screenshots) and I just didn't think the author would have appreciated it.

I generally thought of these plugins more like `pypi` packages. In the end I reckon all plugins scripts are fairly easy to customise for anyone to suit their needs and I also submitted this to have another python plugin example, but I completely understand that it doesn't make sense to have three of the same type.

keithamus:
@chrisidefix it's all cool :sunglasses:, you're right to raise a new PR.

We had a similar scenario before with multiple Pomodoro plugins and we ended up having an issue (https://github.com/matryer/bitbar-plugins/issues/180) to discuss which was best. 

I think the plugins repo is more of a showcase than something like `pypi` or `npm`. Fixes for different problems, over different fixes for the same problems :smile:.

PR #363, opened 2016-03-30
Corrected SI-units and added internet on check

vikeri:
- The SI unit for distance (should be velocity) was wrongly km/h, it should be m/s
- The plugin crashed if there were no internet connection, added check to see if internet is available.


tresni:
Please address the `pyflakes` linter issues.  You have a couple unreferenced variables.  Simply removing the assignment operation should resolve.

vikeri:
Alright! Fixed now :smile:

PR #364, opened 2016-03-30
Transmission remote status plugins

chrisidefix:
Checking transmission status, info and allowing adjustment to alternative speed.
Requires a single initial setup: `python transmission.4m.py init`
This plugin displays all not yet completed torrents using "transmission-remote" cli.
You can pause any torrent in the list by selecting them.
If all active downloads have completed, but some torrents had been paused, the oldest one will be unpaused automatically.

### DIFF
--- a/Lifestyle/transmission.4m.py
+++ b/Lifestyle/transmission.4m.py
@@ -50,17 +50,15 @@ def check_transmission(argv):
     host_pswd = keyring.get_password("transmission-bitbar", host_user)
 
     if 'slowdown' in argv:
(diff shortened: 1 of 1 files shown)


tresni:
Please address the issues raised by pyflakes in the travis-ci build.  Given the variables aren't used, simply don't assign them and you should be good to go ;-)

chrisidefix:
Yeah, thanks - what about python dependencies - is Travis able to handle that automatically?

tresni:
it does static analysis so only cares about syntax and does it "look right". We don't actually see if it runs for that very reason

PR #375, opened 2016-04-07
Mtop

ganeshv:
New plugin: mtop

CPU utilization bar graph is rendered onto a 25x16 BMP file created from scratch with no external dependencies. The dropdown contains current usage, load average and the top 5 CPU-hogs as reported by `top`.

### DIFF
--- a/System/mtop.5s.sh
+++ b/System/mtop.5s.sh
@@ -1,0 +1,217 @@
+#!/usr/bin/env bash
+
+# <bitbar.title>CPU Usage Graph</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Ganesh V</bitbar.author>
+# <bitbar.author.github>ganeshv</bitbar.author.github>
+# <bitbar.desc>CPU usage bar graph</bitbar.desc>
+# <bitbar.image>https://raw.github.com/ganeshv/mtop/master/screenshots/mtop2.png</bitbar.image>
+# <bitbar.about>https://github.com/ganeshv/mtop</bitbar.about>
+
+# CPU utilization bar graph is rendered onto a 25x16 BMP file created from
+# scratch with no external dependencies. The dropdown contains current usage,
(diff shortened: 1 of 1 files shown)


ganeshv:
I did run shellcheck on it, and I think the remaining two warnings are unavoidable,

tresni:
You've managed to get my PR template (that's not finalized) in your branch.  You may want to either "git rebase -i origin/master" and drop d5331d or do something similar :)  You'll have to do a force push at that point to get your branch updated.

As for the warnings Travis Build issues:

1. You are missing metadata, add

    ```sh
    # <bitbar.author></bitbar.author>
    # <bitbar.author.github></bitbar.author.github>
    ```
2. The "Double quote" on line 35 should be doable:

    ```sh
    X"$(find "$HISTORY_FILE" -mtime -2m)"
    ```

3. For line 50, add a line before it that says:

    ```sh
    # shellcheck disable=SC2154
   ```
That will ignore just that error on the next line

keithamus:
For L50 you can actually prevent the error by declaring the variable as itself. Take for example the following:

```bash
#!/bin/bash
foobar="$foobar";
echo "$foobar"
```

This wont error on the `echo "$foobar"` because it is already assigned on L2.

iosdeveloper:
Did you consider removing the `$BitBarDarkMode` check and using `templateImage` with black pixels instead of `image`? That would allow OS X to apply the correct style on highlight.

ganeshv:
@iosdeveloper : I had some idea of using different colors in the graph for user and system CPU, but I think your suggestion makes sense. It looks more native. I've changed it as follows:

  * removed explicit dark mode handling 
  * templateImage for Mavericks onwards
  * image for Mountain Lion (doesn't look nice when clicked if I use templateImage)
  * Those dropdown lines which I want to appear as black instead of the default barely-visible-gray now have a dummy action (refresh) associated with them

@matryer: shellcheck passes now.

iosdeveloper:
@ganeshv 👍 Feel free to go back to `image` once you add colors.

PR #376, opened 2016-04-07
Fix wc output to show count only.

escobarj:
On my systems wc -c file was also outputting the filename with the count and caused the script to fail the numeric compare.


tresni:
That fails shellcheck's linting.  You probably want something like
```
wc -c asdf.lisp | awk ' { print $1 } '
```

PR #378, opened 2016-04-07
Show ✔ instead of ↑0 when there are no updates

SCdF:
Replace the zero count with a tick. Subjective as to whether or not this is better, but I like it more.


tresni:
I think this is an excellent example of a modification that's great to do locally :)  Everyone has their own takes.  (I use 🍺  for homebrew and 🍷  for homebrew-cask personally.)  Given this doesn't change the actual functionality of the plugin, I'd say we leave it for the end-user to decide what they want to display ;-)

PR #380, opened 2016-04-08
Add jira tickets plugin

ryansydnor:
This plugin will display all tickets assigned to the user that
are currently open and group them by status.

### DIFF
--- a/Dev/Jira/jira_tickets.15m.py
+++ b/Dev/Jira/jira_tickets.15m.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# <bitbar.title>Jira</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Ryan Sydnor</bitbar.author>
+# <bitbar.author.github>ryansydnor</bitbar.author.github>
+# <bitbar.desc>Display your outstanding Jira tickets sorted by status.</bitbar.desc>
+# <bitbar.dependencies>python,pip[jira]</bitbar.dependencies>
+
+import sys
+from collections import defaultdict
+
+jira_url = 'https://jira.example.com'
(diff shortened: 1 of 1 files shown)


tresni:
There is already a [JIRA plugin](https://getbitbar.com/plugins/Dev/Jira/jira.1m.sh).  Per @keithamus' comment in #362:

> [T]he plugins repo is more of a showcase than something like `pypi` or `npm`. Fixes for different problems, over different fixes for the same problems :smile:.

Does your plugin offer additional functionality over the existing JIRA plugin?

ryansydnor:
@tresni - fair. A few things:

1. I couldn't get the other plugin to run easily (kept crashing Bitbar completely, actually)
2. I didn't like the fact the other plugin didn't categorize tickets by status. 
3. In Jira, every team has different workflows and potentially custom status field values. The other plugin doesn't account for that kind of thing.
4. There's a bit of "reinventing the wheel" in the other plugin. There's always a tradeoff, but I feel that leveraging a library to handle talking to Jira greatly simplifies the maintenance burden on the bitbar plugin.

Just some thoughts. Happy to close the PR if you don't think they're valid :)

keithamus:
@ryansydnor perhaps you would consider adding features and fixing bugs in the existing Jira plugin?

ryansydnor:
@ngehlert - as the original author, what do you think? Does the tradeoff between maintainability, flexibility, and features at the expense of an extra dependency warrant swapping out your Jira plugin? If not, I'm happy to close this PR.

ngehlert:
@ryansydnor i'll have a look at the new plugin. but since it apparently introduces a new dependency i'm already not a big fan. 
i edit this comment in a couple of minutes

ryansydnor:
Ok! I'll close this in favor of the existing one. 

It sounds like a configuration driven flag could be a great solution to the various grouping mechanisms. Looking forward to config being implemented.

PR #387, opened 2016-04-11
Adding internet check to bitbar version plugin

oliviernt:
prevents showing and error diretly when re-connecting to wifi

### DIFF
--- a/BitBar/version.12h.rb
+++ b/BitBar/version.12h.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 #
 # <bitbar.title>BitBar Version</bitbar.title>
-# <bitbar.version>v0.1.0</bitbar.version>
+# <bitbar.version>v0.2.0</bitbar.version>
 # <bitbar.author>Olivier Tille</bitbar.author>
 # <bitbar.author.github>oliviernt</bitbar.author.github>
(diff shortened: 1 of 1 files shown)


keithamus:
Hmm... would it not be easier to just try/catch on fetching the new version, rather than pinging google.com?

oliviernt:
You're right. I missed the easy change :D

BitBar/version.12h.rb
```diff
@@ -16,6 +16,7 @@
 require 'net/http'
 require 'json'
 require 'nokogiri'
+require 'resolv'
```
keithamus: Is resolv needed any more?

BitBar/version.12h.rb
```diff
@@ -26,8 +27,22 @@
...
+  while json_result.nil?
+    begin
+      json_result = JSON.parse(Net::HTTP.get(url))
+    rescue => e
+      sleep_time *= 2
+      if sleep_time < sleep_until
+        sleep sleep_time
+      else
+        throw e
+      end
```
keithamus: I like the exponential backoff, very clever 👍

BitBar/version.12h.rb
oliviernt: Yup. Missed that.

PR #390, opened 2016-04-18
New plugin: mdf - Disk usage stats

ganeshv:
`mdf` displays a rectangular icon on the Mac OS X menu bar which is filled in proportion to disk space used on the primary disk. Clicking the icon opens a dropdown containing details of disk space utilization for each physical disk.


keithamus:
Wow this is super cool! 🎉

PR #391, opened 2016-04-19
Formula 1

destroy-everything:
Shows countdown to next event FP1,2,3 Qualifying and Race


manojlds:
@destroy-everything - can you do the corrections as shown in the build errors - https://travis-ci.org/matryer/bitbar-plugins/builds/124159769

Issue #373, opened 2016-04-07
Netinfo Plugin


Gram21:
The Netinfo Plugin has a bug in the ipv6 curl adress. 
Line 52 should be changed form
`EXTERNAL_IP6=$(curl --connect-timeout 3 -s http://v6.ipv6-test.com/api/myip.php || echo None)`
to
`EXTERNAL_IP6=$(curl -6 --connect-timeout 3 -s http://v6.ipv6-test.com/api/myip.php || echo None)`

Otherwise the URL cannot be resolved by curl.

q1x:
Seems to be working for me. Nevertheless, it's a good idea to force ipv6.

I'll prepare a pull request.

q1x:
https://github.com/matryer/bitbar-plugins/pull/395

PR #395, opened 2016-04-22
Added -4 and -6 flags to curl commands to only resolve the proper ...

q1x:
protocol. Also removed color for the "Speedtest" menu entry.

This resolves https://github.com/matryer/bitbar-plugins/issues/373


keithamus:
LGTM; perhaps @Gram21 would like to test it to make sure it does indeed fix #373 for them.

Gram21:
Works fine for me, too!

PR #396, opened 2016-04-24
Add a script to get a news feed on Feedly

morizyun:
The existing "feedly.rb" gets a feed on Feedly. I wanted to check tech blog more effectively 🍻 

Example:

Feedly


morizyun:
Thank you very much 💯

PR #399, opened 2016-04-27
Added KillDock plugin

indapublic:
Kill Dock. It useful when Dock is missing on main display


manojlds:
@indapublic you need to make the script executable.

PR #403, opened 2016-04-27
Small fix for redmine plugin.

chilic:
The advantage of #!/usr/bin/env ruby is that it will use whatever ruby executable appears first in the user's $PATH.


tresni:
The problem with `/usr/bin/env` is that due to known issues post OS X 10.10 `$PATH` is not always what you expect it to be for applications launched via launchd (most applications in OSX.)  So, a lot of times, using /usr/bin/env for a script within a running application will lead to using system install of whatever vs Homebrew/MacPorts/rbenv/etc.

But, it's probably still safer then forcing the use of Homebrew installed ruby which not everyone may have ;-)

PR #407, opened 2016-05-03
[WIP] Change desktop wallpaper via reddit

milch:
This is still a WIP, I'd just like to get some input before I continue. Right now, there are no menus except for the menu bar logo (also, errors are shown in the dropdown).

* I had to split up the project into multiple files, are there any 'best practices' for BitBar plugins with more than one file yet? I simply put all the project files in a subfolder and used a simple script to call my main file in that folder, and BitBar seems to deal with it fine. Is this okay? 
* I'd like to have just a small amount of configurable options for this plugin, likely just the list of subreddits to fetch the wallpapers from, the sort and timeframe options, and stuff like that. Is there some kind of standardised way to do some simple configuration like that? I don't know if coding all the configuration stuff into the plugin would make sense, since more plugins could benefit from having a standardised solution for simple configuration options (if it doesn't already exist).
* It seems that the menu bar icon can not have retina resolution. It'd be great if there was some way to set a retina icon since the 16x16 standard icon doesn't look very great on retina screens.


iosdeveloper:
Hi Manu,

* Our [URL scheme](https://github.com/matryer/bitbar/blob/master/Docs/URLScheme.md#openplugin) we use on getbitbar.com can only handle a single file at the moment. Plugin bundles are still in [discussion](https://github.com/matryer/bitbar/issues/112). Maybe we should change it to clone the plugins repo and create symbolic links in the [`Enabled` folder](https://github.com/matryer/bitbar#using-symlinks).
* Same for [configuration](https://github.com/matryer/bitbar/issues/88).
* Don't double the size, double the resolution for `@2x` images: https://github.com/matryer/bitbar/issues/314

tresni:
As noted by @iosdeveloper we can't support multi-file plugins at this time.  I'm closing this for the time being.  Please feel free to reopen if/when plugin bundles are supported or this is a single file plugin.

PR #412, opened 2016-05-08
Vagrant Plugin Fixes

bbatsche:


axeloz:
Thanks for the cc. 
I didn't think of the spaces in path (who uses spaces in tabs anyways? ;-) )

PR #413, opened 2016-05-09
Extended battery level indicator plugin

alexrockt:
shows battery status for trackpad, keyboard and laptop. Based on the trackpad-dot plugin by @milch


milch:
Looks nice, and thanks for giving me credit for the original plugin!

Issue #415, opened 2016-05-13
re: Hacker News


rkumar:
FAO: (@manorie)
I get this on running Hacker News.

/Users/xxx/bitbar/hacker_news.1m.rb:32:in `initialize': No such file or directory @ rb_sysopen - /Users/xxx/hacker_news_data.txt (Errno::ENOENT)
	from /Users/xxx/bitbar/hacker_news.1m.rb:32:in `open'

I tried `touch` to create the file, but then got a `marshal data too short`. So how does one run this the first time, when the file does not exist ?

Thanks.

manorie:
Thanks for reporting @rkumar ,
I have fixed the problem will be sending PR asap.

rkumar:
Thanks, it is working great now.

manorie:
you are welcome @rkumar

PR #417, opened 2016-05-13
#415 - HackerNews plugin, fix file error and sublinks.

manorie:
I just made some fixes to the plugin post issue #415 , plus I am sending a minor improvement of adding sublinks for news comments.
@keithamus @caneroj1 @thameera @oliviernt


keithamus:
👍

Issue #416, opened 2016-05-13
re: Subreddit Links


rkumar:
FAO: (@caneroj1)

When I run this, I get no data returned. The hash "data" has only one key namely "error". When i run the same URL on my browser, I do get JSON data, but using this script, and also if I run `curl https://www.reddit.com//top.json` on the command line using curl, I get "{"error": 429}% ".

caneroj1:
I just opened pull request #418 to take care of this issue.

Error 429 was being received because our user-agent was being rate-limited. The change solves the problem for me. Let me know if it still doesn't work.

rkumar:
Works fine, now. Thanks a lot.

PR #418, opened 2016-05-14
updated subreddit links to specify a user agent.

caneroj1:
This pull request addresses issue #416, where the request to get subreddit links was receiving an HTTP 429 Error. 

That error was being received because reddit was rate-limiting our request when it was coming in from the default user-agent Ruby uses in the Net/HTTP library. We now specify a generic bitbar user-agent when we send the request. That user-agent can also be changed on a per user basis, so people can add a more detailed user-agent if they want.


tresni:
Please correct the rubocop linking error reported by Travis and we'll get this merged. Looks like you are assigning to a variable you never use.

caneroj1:
My bad! Should be taken care of now.

PR #422, opened 2016-05-18
Add Stock Portfolio plugin

lukstei:


keithamus:
Hey @lukstei. This is a nice looking stock plugin, but we already have 2 (https://github.com/matryer/bitbar-plugins/blob/master/Finance/stock.5s.sh and https://github.com/matryer/bitbar-plugins/blob/master/Finance/yahoo-stock.1m.sh). What makes this one different?

lukstei:
Hey @keithamus,
it shows the cumulated profit of your portfolio, thats what makes it interesting.
The others only show individual stocks but usually one cares about the whole portfolio (sum of short+long), thats why I created it.

keithamus:
@lukstei perhaps a name to better indicate this would be good `stock-portfolio-pl` maybe?

lukstei:
Good point, I'll change it

PR #423, opened 2016-05-18
Add check for update count

brandonpittman:
If there aren't any updates, there's no reason to clutter the menu bar.

### DIFF
--- a/System/mas.1d.sh
+++ b/System/mas.1d.sh
@@ -13,9 +13,11 @@ UPDATES=$(/usr/local/bin/mas outdated);
 
 UPDATE_COUNT=$(echo "$UPDATES" | grep -c '[^[:space:]]');
 
-echo "↑$UPDATE_COUNT | dropdown=false"
-echo "---";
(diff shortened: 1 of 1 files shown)


System/mas.1d.sh
```diff
@@ -9,13 +9,19 @@
 # <bitbar.dependencies>mas</bitbar.dependencies>
 # <bitbar.abouturl>https://github.com/matryer/bitbar-plugins/blob/master/System/mas.1d.sh</bitbar.abouturl>
 
-UPDATES=$(/usr/local/bin/mas outdated);
+if test $UPDATE_COUNT -gt 0; then
```
tresni: I think you jumped to far ahead here.  This doesn't test for `mas` this checks that somehow the variable `UPDATE_COUNT` is set and not 0.

`if [ -n "$(which mas)" ]` is probably closer to what you are looking for to see if `mas` is installed.

brandonpittman:
Opened different PR with update code

PR #424, opened 2016-05-19
Add check for `mas` install and update count

brandonpittman:
I think this is what I intended originally.


tresni:
Please address the [linting issues](https://travis-ci.org/matryer/bitbar-plugins/builds/131303135#L345-L353) from shellcheck and we'll get this merged :)

PR #428, opened 2016-05-20
New plugin: CrossFit regionals results.

angrytongan:


tresni:
Hi @angrytongan and thanks for the PR. We lint all plugins against a standard language linter, for shell scripts that is shellcheck. Please review and handle the linting errors in the linked Travis build and we'll get this merged!

PR #430, opened 2016-05-24
update todoist.1m.py

chiisaihayashi:
todoist.1m.py did not handle recurring task.
To fix this problem, some lines were changed.


chiisaihayashi:
@keithamus
I tested 3 types of recurring tasks ( every day, every week, every month).
This works well for all of them. Attached is a sample executed on my laptop.
screen shot 2016-05-26 at 13 34 55

douglance:
Issue is resolved.

PR #435, opened 2016-06-04
Update version

tanrax:
Best performance and port to python


tresni:
Are you looking to replace the previous .sh script with this python script?  The current commits (which seem to combine the old and new version) don't do that.  From what I'm seeing, this would simply create a new battery health script.

tanrax:
Yes, my intention is to replace the .sh for a python version. Excuse my English, I do not understand the problem :)

tresni:
I don't think this will get rid of the shell version of the plugin, but we'll give it a shot :)

PR #439, opened 2016-06-09
Add Lisk/BTC Poloniex plugin

corv89:


tresni:
Please correct the following linter error and commit the change:

```
In /Users/travis/build/matryer/bitbar-plugins/Finance/lskbtc.sh line 12:
echo -n "LSK "; curl -s https://poloniex.com/public\?command\=returnTicker | grep -Eo '"BTC_LSK":{"id":163,"last":"[0-9]+\.[0-9]+' | sed 's/"BTC_LSK":{"id":163,"last":"//'
                                                            ^-- SC1001: This \= will be a regular '=' in this context.
```

Should just need to remove the backslash here as `=` does not need escaped.

matryer:
@corv89 please correct that error and we get get this merged in :)

corv89:
Done

PR #440, opened 2016-06-10
Adobe Flash Update Check, PIA VPN Status & Currency Rates

parvez:


tresni:
In the future, generally best to do 1 PR per plugin.  That way we can merge things individually if there are issues (like the PIA VPN plugin failing.)

parvez:
Sure! Thanks!

PR #441, opened 2016-06-12
Filter out extra annotations from tools such as bugwarrior

### LINKED ISSUE #421: re: Taskwarrior
FAO: (@christophruss)
Is there a way to have it show the list with the annotations filtered out? The extra annotations from taskwarrior are making the list quite long. I looked at the code myself, but I'm pretty new to taskwarrior so I'm having difficulty understanding the correct filters I need to use for the 'next' or 'completed' list commands.

kfdm:
Fixes #421

This should go through and parse out the extra empty lines and only look
for lines that start with a digit (task ID) or one of the markers used
to print up the UI

Will need to have @chrisidefix check it since he is the original author of the plugin.
Thanks again for the tip and for your patience in my delayed responses :)

### DIFF
--- a/Lifestyle/taskwarrior.4m.py
+++ b/Lifestyle/taskwarrior.4m.py
@@ -65,7 +65,7 @@ def print_output(cmd,color,head,print_content=True,command='',ignore_id_list=[],
         # When looking for 'active' or 'next' tasks, we want to look only at
         # lines that start with a digit or a -. Other line are likely extra
         # data like annotations from bugwarrior sync
-        content_id = output_line.split()[0].strip()
+        content_id = output_line.split()[0]
         if not content_id.isdigit() and content_id not in ['--', '-', 'ID']:
             continue
         content_lines.append(output_line)
--- a/Lifestyle/taskwarrior.4m.py
+++ b/Lifestyle/taskwarrior.4m.py
@@ -60,8 +60,15 @@ def print_output(cmd,color,head,print_content=True,command='',ignore_id_list=[],
 
     for output_line in output_lines:
         output_line = output_line.strip()
-        if len(output_line) > 0:
-            content_lines.append(output_line)
+        if not output_line:
+            continue
+        # When looking for 'active' or 'next' tasks, we want to look only at
+        # lines that start with a digit or a -. Other line are likely extra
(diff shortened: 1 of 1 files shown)


Lifestyle/taskwarrior.4m.py
```diff
@@ -60,8 +60,15 @@ def print_output(cmd,color,head,print_content=True,command='',ignore_id_list=[],
...
     for output_line in output_lines:
         output_line = output_line.strip()
-        if len(output_line) > 0:
-            content_lines.append(output_line)
+        if not output_line:
+            continue
+        # When looking for 'active' or 'next' tasks, we want to look only at
+        # lines that start with a digit or a -. Other line are likely extra
+        # data like annotations from bugwarrior sync
+        content_id = output_line.split()[0].strip()
```
chrisidefix: `.strip()` is probably not required here

Lifestyle/taskwarrior.4m.py
```diff
@@ -60,8 +60,15 @@ def print_output(cmd,color,head,print_content=True,command='',ignore_id_list=[],
...
+        if not content_id.isdigit() and content_id not in ['--', '-', 'ID']:
```
chrisidefix: (1) do you think it would actually make more sense to check (before line 62) if `output_line` starts with multiple `space`s ?

(2) what about an absolute max length of each line - and just replace the remaining bits with `...` ?

Lifestyle/taskwarrior.4m.py
kfdm: You're right. This is probably just from my testing and then not cleaning up after I was done. Good catch.

Lifestyle/taskwarrior.4m.py
kfdm: (1) I'm not sure if it makes so much difference. I don't think one needs to worry so much about making this plugin super optimized. That could be left for someone else that is more picky about things.

(2) I think it already trims the line length and replaces with `...`. My problem at the beginning was that the extra annotation data from [bugwarrior](http://bugwarrior.readthedocs.io/en/latest/) made the list way too long with a lot of unnecessary data.

PR #442, opened 2016-06-12
Salyangoz.me Recent Shared BitBar Plugin

s4l1h:


tresni:
Currently a plugin should be a single file.  We don't have a way to support multiple files and it makes it confusing to the user.  Also, if go can not be interpreted directly and must be compiled, then it's not a good candidate for inclusion in this repo at this time.

tresni:
Given no response in over a month I'm going to close this PR.  Please feel free to reopen if/when these issues are addressed.

PR #444, opened 2016-06-18
[Emacs Package Updates] Rescue melpa.org returns nil

masutaka:
Sometimes melpa.org returns nil. This Pull Request rescues it.


masutaka:
Thanks a lot!

PR #450, opened 2016-06-22
Adding Checkman Simulator plugin

deluan:
This plugin simulates Checkman functionality, allowing you to use all its
plugins and creating checks in external files. It downloads Checkman's plugins
on demand, so you don't need to install Checkman itself.

To learn more about Checkman, and learn about its existing plugins, see:
https://github.com/cppforlife/checkman#included-check-scripts

Configuration: Same as for Checkman, check https://github.com/cppforlife/checkman#configuring-checkman-via-checkfiles

Missing/Not yet implemented:
- Notifications/Stickies

### DIFF
--- a/Dev/Checkman/checkman.10s.rb
+++ b/Dev/Checkman/checkman.10s.rb
@@ -1,0 +1,117 @@
+#!/usr/bin/env ruby
+
+# <bitbar.title>Checkman Simulator</bitbar.title>
+# <bitbar.version>v0.1</bitbar.version>
+# <bitbar.author>Deluan Quintao</bitbar.author>
+# <bitbar.author.github>deluan</bitbar.author.github>
+# <bitbar.desc>This plugin reuses Checkman's configurations and plugins, so you don't need to install it :)</bitbar.desc>
+# <bitbar.image>http://i.imgur.com/irmlsOX.jpg</bitbar.image>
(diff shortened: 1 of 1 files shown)


tresni:
This is really cool.  Are you wanting this to be merged now or are you still making changes?

deluan:
It is not working with dark menu bar. I'll do more fixes tomorrow and will let you know when it is done.

deluan:
@tresni It is good enough for a first release now. Thanks.

Dev/Checkman/checkman.10s.rb
```diff
@@ -49,9 +49,26 @@
   },
 }
 
+PRIORITY_ORDER = [ICON_OK, ICON_OK_CHANGING, ICON_FAIL, ICON_FAIL_CHANGING, ICON_UNDETERMINED, ICON_UNDETERMINED_CHANGING]
+
+# TODO: What is the parameter to prevent the line to be dimmed? This is
+# the only way I could figure out
+# NO_DIM=" bash=/bin/true terminal=false "
+NO_DIM=" bash=/bin/true terminal=false "
```
tresni: This is totally an OSX thing.  Your workaround is fine.  If you don't want it dimmed it must be an action or submenu.  The other option would be to set the `color=black` but that's not friendly to darkmode...

PR #455, opened 2016-06-25
New Plugin: Get today's steps and sleep duration from the Jawbone UP API

dettmering:
As the title says, this plugin will retrieve today's steps and sleep duration from the Jawbone UP API. You need to retrieve an access token as described here: https://jawbone.com/up/developer/authentication


matryer:
Excellent.

PR #466, opened 2016-07-05
Add a generic package manager plugin.

kdeldycke:
Currently supports Homebrew and Cask.

I plan to add others like Python's `pip` in future revisions.

PR #469, opened 2016-07-06
Expose Python package updates via Pip

kdeldycke:
Following #466, here is an update exposing latest Python packages updates.


kdeldycke:
Thanks @keithamus for the merge! :)

PR #480, opened 2016-07-09
Fix a couple issues with npm/apm parsing

tresni:
FAO: @kdeldycke

A couple tweaks after investigating what @keithamus brought up with linked packages.

### DIFF
--- a/Dev/Homebrew/package_manager.7h.py
+++ b/Dev/Homebrew/package_manager.7h.py
@@ -424,15 +424,22 @@ class APM(PackageManager):
 
 
 class Gems(PackageManager):
+    HOMEBREW_PATH = '/usr/local/bin/gem'
+    SYSTEM_PATH = '/usr/bin/gem'
+
+    def __init__(self):
+        super(Gems, self).__init__()
+
+        self.system = True
+        if os.path.exists(Gems.HOMEBREW_PATH):
+            self.system = False
+            self._cli = Gems.HOMEBREW_PATH
+        else:
+            self._cli = Gems.SYSTEM_PATH
 
     @property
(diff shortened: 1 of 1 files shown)


Dev/Homebrew/package_manager.7h.py
```diff
@@ -214,7 +214,12 @@ def name(self):
     def sync(self):
         output = self.run(self.cli, '-g', '--progress=false', '--json',
                           'outdated')
+        if not output:
+            return
+
         for package, values in json.loads(output).iteritems():
+            if values['current'] == 'link':
```
keithamus: needs to be `values['wanted'] == 'linked'`. I know this seems very counter intuitive, but the `current` is the version that is in the package.json, which is likely never `linked`.

tresni:
gotta update for @kdeldycke's update anyway so I'll get his (and a `sudo` on the gem commands)

tresni:
rebased on latest master and using wanted instead of installed.

tresni:
Added an update for Gems support when using system ruby.

Dev/Homebrew/package_manager.7h.py
```diff
@@ -371,7 +371,12 @@ def sync(self):
         """
         output = self.run(
             self.cli, '-g', '--progress=false', '--json', 'outdated')
+        if not output:
+            return
+
         for package, values in json.loads(output).iteritems():
+            if values['wanted'] == 'link':
```
keithamus: `'linked'` not `'link'` btw.

kdeldycke:
Thanks @tresni for the follow-up! :)

PR #483, opened 2016-07-10
Add plugin AKAlias.

irstacks:


irstacks:
Shellcheck had some good suggestions. But now the robot is wrong. I think.

Dev/Shell/aka.1h.sh
```diff
@@ -0,0 +1,75 @@
...
+# 
+# Header. 
+echo "= | color=#1853C9"  # ∷ ⚯ ɐ ª
+echo '---'
+
+# Read each line in aka file. 
+while read -r line 
+do 
+	# Strip 'alias ' from line for legibility. 
+	dict="$(echo ${line//alias /})"
```
keithamus: Its not wrong, this could be `dict="${line//alias /}"`

tresni:
@irstacks If you are okay with what you have you can suppress those warnings.  Just add ` # shellcheck disable=SC2116,SC2086` to the line before.

I will say I don't think shellcheck is wrong on any of those:
* `echo "$(echo $variable)"` is the same as `echo "$variable"`.
* as @keithamus pointed out, for assignment, you can safely just do the variable substitution without need to do the `echo`.

Doesn't make these "wrong", just maybe not optimal :)

tresni:
Not sure why Travis isn't picking up that commit...

PR #486, opened 2016-07-12
Plugin akalias

irstacks:
Fixes issue I had where better (more efficient) syntax lead to all lines not printing.  Don't understand it.


tresni:
Also not rebased properly, but given this is just one plugin, I think I can safely merge this one.  Please see my comment on your other plugin about proper rebasing.

PR #505, opened 2016-07-29
Updates the artist/song text immediately when you move tracks

joshlove:
Pretty simple change. When you move forward/back a track the name of the artist and song only updates when the script runs. This will refresh it upon moving tracks, updating that information immediately.


tresni:
Generally I'd recommend incrementing the release version, but we've never been super picky about that ;-)

PR #510, opened 2016-08-03
add daily summary to ForecastIO/weather

blockloop:
image

### DIFF
--- a/Weather/ForecastIO/weather.15m.py
+++ b/Weather/ForecastIO/weather.15m.py
@@ -11,6 +11,7 @@
 
 import json
 import urllib2
+import textwrap
 import base64
 from random import randint
 
@@ -142,6 +143,11 @@ def get_wx():
         if item == 'summary':
           weather_data['next_hour'] = str((wx['minutely']['summary'].encode('utf-8')))
 
+    if 'daily' in wx:
+      for item in wx['daily']:
+        if item == 'summary':
+          weather_data['week'] = str((wx['daily']['summary'].encode('utf-8', 'ignore')))
+
     if 'city' in location and 'region' in location:
       if location['city'] == '' and location['region'] == '':
         if 'country' in location:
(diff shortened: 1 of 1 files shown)


Weather/ForecastIO/weather.15m.py
```diff
@@ -201,6 +207,10 @@ def render_wx():
 
   print '---'
 
+  if 'pressure' in weather_data:
```
seripap: This is pulling the wrong key value. I would also wrap this below the radar image and keep the next hour summary at the top as it's the most relevant.

Something like this:

```
  print '---'

  if 'next_hour' in weather_data:
    print weather_data['next_hour']

  print '---'

  if 'image' in weather_data:
    if weather_data['image']['encoded']:
      print '| href=' + weather_data['image']['url'] + '| image=' + weather_data['image']['encoded'] 

  print '---'

  if 'week' in weather_data:
    print "\n".join(textwrap.wrap(weather_data['week'], 50))
     print '---'
```

seripap:
@blockloop could you fix from the reference in the diff? Should be good afterwards. Thanks!

blockloop:
oops :) fixed it

Weather/ForecastIO/weather.15m.py
```diff
@@ -201,6 +207,10 @@ def render_wx():
 
   print '---'
 
+  if 'week' in weather_data:
+    print "\n".join(textwrap.wrap(weather_data['week'], 50))
+    print '---'
+
   if 'next_hour' in weather_data:
```
seripap: Can you move this block to line 203, above the image and add another line break? This information is more relevant versus the weekly summary so it should naturally flow as you're reading

blockloop: Like this?

image

blockloop:
I moved the daily summary above the image

seripap:
@blockloop the current summary should be up there- "Clear for the hour'. Swap the with the daily summary

blockloop:
I moved it, but it wasn't there before I made my changes. Hourly was after the image.

seripap:
Great- with the addition of daily, it made sense to move it above the image; since the image shows what is actually going on in terms of precipitation for the next hour. 

@keithamus good to merge 👍

tresni:
Thanks for the discussion and PR!

PR #514, opened 2016-08-06
Show lock/unlock icons for private/public transfers

rchiechi:


rchiechi:
The Travis CI checker is parsing with python2 and (incorrectly) reporting a syntax error with print('string', end='').

tresni:
Oi vey.  Good catch.  I'll update [.test.py](.test.py) to deal with python2 vs python3.

tresni:
I believe we are just waiting on #515 and then we can re-run the test and get this merged.

tresni:
hmm.. Not getting the latest travis.yml configuration.  Going to try closing/reopening to see if that does it..

tresni:
That seemed to grab the right config so hopefully this will test properly :)

PR #515, opened 2016-08-09
Multi linter

tresni:
Supports multiple linters for the same file extensions.  Useful where the file extension may be shared, but the linter should be different (e.g. python2 vs python3.)

This should allow #514 to pass build tests.


tresni:
@keithamus any opinion?

keithamus:
Ultimately LGTM. I'm curious though; it seems like it supports multiple linters running on the same file? Unlikely to be a problem I'm sure, but could be fun watching 2 linters fighting it out.

tresni:
Yeah, it does.  Since (e.g.) python could mean python2 or python3 depending on how you setup your computer.  So in that case it just looks to see if 1 of them says the file is good :)  Best answer I could come up with today.

PR #516, opened 2016-08-10
Update Package Manager plugin to v1.6

### LINKED ISSUE #493: re: Homebrew Cask Updater and Homebrew Updates
FAO: (@srdgh)
FAO: (@killercup)

I am using =these two plugins together, and it seems that when I am doing a Refresh All one of the two fails as both are trying to do a `brew update` which is not possible.

Would it be possible to add in one (or both) of these plugins a condition which waits for the `brew update` to finish and than continues?

Thanks,

Rainer

kdeldycke:
Changelog:
* Work around the lacks of full pip upgrade command.
* Fix UnicodeDecodeError on parsing CLI output (closes #493).

### DIFF
--- a/Dev/Homebrew/package_manager.7h.py
+++ b/Dev/Homebrew/package_manager.7h.py
@@ -110,6 +110,7 @@ class PackageManager(object):
     def run(self, *args):
         """ Run a shell command, return the output and keep error message.
         """
+        self.error = None
         process = Popen(
             args, stdout=PIPE, stderr=PIPE, universal_newlines=True)
         output, error = process.communicate()
--- a/Dev/Homebrew/package_manager.7h.py
+++ b/Dev/Homebrew/package_manager.7h.py
@@ -1,28 +1,45 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # <bitbar.title>Package Manager</bitbar.title>
-# <bitbar.version>v1.5</bitbar.version>
+# <bitbar.version>v1.6</bitbar.version>
 # <bitbar.author>Kevin Deldycke</bitbar.author>
 # <bitbar.author.github>kdeldycke</bitbar.author.github>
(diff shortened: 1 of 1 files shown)


Dev/Homebrew/package_manager.7h.py
```diff
@@ -93,12 +110,12 @@ def active(self):
     def run(self, *args):
         """ Run a shell command, return the output and keep error message.
         """
-        self.error = None
```
tresni: Pretty sure you want to leave this as nothing else clears the error state. So without this don't you end up with a single error never going away? (Mobile so maybe I'm missing a similar line somewhere else.)

kdeldycke: Oh you're right. Overlooked this change. Going to fix it right away.

kdeldycke: Fixed in f27306433545501e2a050afa46924f70596ed19c .

kdeldycke:
Thanks @tresni for the merge! :)

tresni:
Sadly this fails if you have more then a handful of packages that need updated.  The issue is that BitBar only accepts [5 params](https://github.com/matryer/bitbar/blob/master/App/BitBar/Plugin.m#L220-L224).  In my case, I had 20 packages that needed updated.  It'll update the first 3 (since `install` and `--upgrade` take up `params1=` and `params2=`.)

I have an idea for a potential way to address this, hopefully have something done tonight.

kdeldycke:
Thanks for the hint. Didn't knew about this limitation, but suspected it.

Pip full upgrade is still a hack. Having the upgrade pass limited to 5 items is good enough to me. It's just slightly inconvenient as you have to call it a couple of time to flush the whole queue. But the plugin is there to remind you about the left-over upgrades.

Good if you find a better workaround but that's not a high-priority task.

tresni:
I think I have a decent work around that will allow us to support any package manager that doesn't normally support a full update. PR should be coming this weekend, but I am traveling so may be delayed ;)

-- Brian

PR #520, opened 2016-08-10
Package Manager updates

tresni:
Fixes a weird NPM issue where modules aren't actually there but it thinks they need to be
Move to out of Homebrew folder as this is not just Homebrew.


tresni:
A couple updates for @kdeldycke to review :)

kdeldycke:
LGTM. Let's merge! :)

PR #522, opened 2016-08-11
Add /usr/local/bin to path if it is missing.

torkelsson:
GUI apps on OS X usually don’t have /usr/local/bin in their PATH by default, so add it if it is missing. Needed for at least npm, which does a /usr/bin/env node to find out where node is installed. 

See for instance the Homebrew FAQ, https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/FAQ.md#my-mac-apps-dont-find-usrlocalbin-utilities

### DIFF
--- a/Dev/package_manager.7h.py
+++ b/Dev/package_manager.7h.py
@@ -88,6 +88,10 @@ class PackageManager(object):
 
     cli = None
 
+    path_list = os.environ.get('PATH','').split(os.pathsep)
+    if not '/usr/local/bin' in path_list:
+        os.environ['PATH'] += ':/usr/local/bin'
+
     def __init__(self):
         # List all available updates and their versions.
         self.updates = []
--- a/Dev/package_manager.7h.py
+++ b/Dev/package_manager.7h.py
@@ -88,9 +88,26 @@ class PackageManager(object):
 
     cli = None
 
-    path_list = os.environ.get('PATH','').split(os.pathsep)
-    if not '/usr/local/bin' in path_list:
-        os.environ['PATH'] += ':/usr/local/bin'
+    # OS X does not put /usr/local/bin or /opt/local/bin in
+    # the PATH for GUI apps. For some package managers this
+    # is a problem. If you get an error similar to
+    #
+    # "env: node: No such file or directory
+    #
+    # it is most likely that something is missing from the
+    # path. However Homebrew and Macports are using different pathes.
+    #
+    # For Homebrew uncomment next line:
+    #extra_path='/usr/local/bin'
+    # For MacPorts uncomment next line
+    #extra_path='/opt/local/bin'
+    # For both Homebrew and MacPorts uncomment the following lines:
+    #path_list = os.environ.get('PATH','').split(os.pathsep)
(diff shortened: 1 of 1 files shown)


tresni:
This is left up to the user as (e.g.) MacPorts installs things in /opt/ where Homebrew installs in /usr/local/bin

iosdeveloper:
Relates matryer/bitbar#353

plessbd:
Should the file be updated with this change in the changelog as well?

torkelsson:
Change the PR to include comments to package_manager how to enable /usr/local/bin or /opt/local/bin instead. With commented out code.

Dev/package_manager.7h.py
```diff
@@ -88,9 +88,26 @@ class PackageManager(object):
...
+    # For Homebrew uncomment next line:
+    #extra_path='/usr/local/bin'
+    # For MacPorts uncomment next line
+    #extra_path='/opt/local/bin'
+    # For both Homebrew and MacPorts uncomment the following lines:
+    #path_list = os.environ.get('PATH','').split(os.pathsep)
+    #if not extra_path in path_list:
+    #    os.environ['PATH'] += ":" + extra_path
+
+    print(os.environ['PATH'])
```
plessbd: This is going to completely break the script

I would also recommend putting anything that the user should / has to modify at the top of the script instead of 90 + lines down.  Most people are not going to look for a variable to change that far down.

Is there any reason to not just add both of these known paths to the PATH instead of even making it optional?

tresni: Agree, if we are going to do this, just overload the path.  It won't hurt anything to have a path appear twice in the `$PATH` itself

plessbd:
The comments here to add the path seem wrong.

They say to uncomment one of the extra path variable assignments.

But doing that wont do anything if you don't uncomment the if conditional as well.

You also cannot do both at the same time currently

I think the best solution would be to either put them in no matter what, or just check for both directories existence and then put them in.

torkelsson:
@plessbd The comments are maybe not wrong but incomplete :-) I agree that they can be improved and perhaps parts of the should be moved to the description of the plugin. And/or the code can be improved so that you only have to uncomment the extra path assignment. 

Is is common that people have both Homebrew and Macports enabled? 

I'm somewhat old school in that I dislike to have the same paths multiple times in PATH, it both clutters PATH and adds (very small) extra search time when looking for binaries. In this case it shouldn't really matter though as the default PATH in OS X is quite short. 

A more philosophical question is if the added paths should be appended or prepended to PATH?

I'm also unsure from the comments if the consensus is that the extra paths should be enabled by default or if it is left as an exercise for the user to enable them.

plessbd:
I have known a few people that have had both macports and homebrew, however I do not.

I would probably have an option the user can uncomment and then have the script do two quick checks for the directories(keeping the path to only what actually exists) and prepend to the path.

But I think the updates in #523 take care of this now

torkelsson:
Yeah, I agree that #523 superseds this PR, so I will close this on and turn to some other ideas I have for package_magager.

PR #523, opened 2016-08-13
"Update All" for the rest of us

tresni:
This adds a new feature allowing "Update All" to work with any package manager.
See the implementation for Pip and Homebrew Cask for how I've done it.  Not the
only possible options, but worked well enough for me.

Additionally, a few cosmetic changes and a fix for #522 that should "just work"
for anyone.

### DIFF
--- a/Dev/package_manager.7h.py
+++ b/Dev/package_manager.7h.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # <bitbar.title>Package Manager</bitbar.title>
-# <bitbar.version>v1.6</bitbar.version>
+# <bitbar.version>v1.7</bitbar.version>
 # <bitbar.author>Kevin Deldycke</bitbar.author>
 # <bitbar.author.github>kdeldycke</bitbar.author.github>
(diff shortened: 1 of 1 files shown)


tresni:
Forgot to @kdeldycke so he could review :)

kdeldycke:
Good work!

I propose to reuse @torkelsson comment from https://github.com/matryer/bitbar-plugins/pull/522/files above the `os.environ['PATH']` line, to document the necessity of hacking the `PATH` environment variable.

Also feel free to bump version of the script to 1.7 and add a changelog entry.

That being said, LGTM.

tresni:
The benefit is now PATH ha king is required. I just front load it into the PATH and it'll work for everyone (hopefully) no changes required ;)

-- Brian

torkelsson:
@tresni I think @kdeldycke is suggesting that you add the part of my comment that says why you have to modify PATH at all. Not the part about what the user should do, as that is not necessary with your changes.

kdeldycke:
Yep, @torkelsson got it! :)

tresni:
(doh) that makes total sense.  Adding it now and updating this PR.

tresni:
Once the tests go green, I'll get this merged in unless someone has anything else to add at this point :)

keithamus:
This looks great btw! Getting a scarily good package manager manager tucked away in bitbar now.

Merging this because @tresni was going to when it went green.

tresni:
I pushed a commit (404f1e4) as I introduced a bug where the output from `_update_all_cmd` was being passed through `bitbar_cli_format` twice...

PR #527, opened 2016-08-18
Plugin contribution

PHNTXX:
Hi,
I wrote a little plugin for your app that allows pebble developers to see how many hearts their app has in the pebble app store. I would really appreciate it if you could add it to the main repo. In case there's anything I should change just let me know.


tresni:
Make sure to `chmod +x` your script and address the pyflakes errors in the Travis CI log and we'll get this merged!

PHNTXX:
I'll make the file executable. However, I can't fix the pyflakes errors as the user needs to enter variables and I've made it so that the space for the variables is commented out and should be uncommented when the variable is defined. Hope that makes sense.

PR #530, opened 2016-08-25
Modified bandwidth-tester with minified output

rangedsp:
Menu space is hard to come by on 13" machines. 
So reduced the length of output via super dumb string replace.


tresni:
Had a similar idea, mine looks like this now (still tweaking)

bandwidth-tester

tresni:
You need to change your script from using `/bin/sh` to using `/bin/bash`.  String replacement isn't defined behavior in `sh`.

rangedsp:
Thanks @tresni, my bad, tested it in a separate file.
Your one is pretty cool, I'm just doing what I can in 10min so it can fit on my 13" mbp :)

PR #533, opened 2016-08-29
Fixed path error in pomodoro script

rael9:
Fixed the pomodoro script when the path to the script has special characters.


rael9:
Sorry, I pushed out another script I'm working on before this was done. If you want me to do another PR to undo this I can.

tresni:
With the revert, we should be okay :)

PR #534, opened 2016-08-29
Created a new version of the Docker status plugin

rael9:
Created a new version of the Docker status plugin to support the new Docker for mac.


tresni:
`chmod +x` your script and then it should pass tests.  Is it possible to extend the existing script vs a new one or is there something fundamentally different?

rael9:
Good question! I edited it to work for me and thought I'd share with others. Let me investigate whether I can set it up to work for either version.

rael9:
Ok. I have removed the new version of the script, and edited the old one so that it should work for both the old style docker machines and the newer docker VM.

tresni:
LGTM!

PR #540, opened 2016-09-01
tidied up structure

matryer:
Moved things around a little, will break some URLs on getbitbar.com sadly.


matryer:
@keithamus

iosdeveloper:
👍

PR #542, opened 2016-09-01
Create ad-pass-expiry

gsprague:
Check out my Active Directory password expiration check. It provides expiration in count of days.


gsprague:
I give up! $((..)) or [[..]] in arithmetic are not passing shellcheck when it's telling me to use those.

tresni:
Please either add an image or remove the `<bitbar.image>` metadata.  Don't think the website is going to handle the `NA` very well.

PR #546, opened 2016-09-04
merge master and PHNTXX-pebble

PHNTXX:


tresni:
I think I got this with the other branch (squashed commits so won't have too crazy history :) )

PR #547, opened 2016-09-04
Plugin Contribution - ISS Pass Finder

BrokenFlows:
A script I adapted to display the upcoming passes of the International
Space Station over your location. Wasn’t quite sure where it belonged
but put it in lifestyle.


BrokenFlows:
I went in and made all the suggested changes but while the check says my use of cat is useless if I use the suggested alternatives my code breaks. What do I do? Do I need to make a new pull request with a new changed file as well?

PR #551, opened 2016-09-08
Update ad-pass-expiry.sh

gsprague:
setting domain var to "yourDomain"


tresni:
Small change, going to go ahead and merge since I'm here.

PR #555, opened 2016-09-09
Add plugin timezone+

aaronedell:
Adding the plugin timezone plus


aaronedell:
This is the error " /Users/travis/build/matryer/bitbar-plugins/Time/timezonesplus.30s.sh image metadata has bad content type: text/html; charset=utf-8"

What does that mean? Its just a URL like all the other plugins have.

iosdeveloper:
If you look closely, other plugins use direct Imgur URLs for the image: http://i.imgur.com/[Image ID].png

PR #557, opened 2016-09-13
add SpursBar

coreyk:
Shows upcoming and previous Tottenham Hotspur match info and headlines


matryer:
Hahahaha - great name

PR #562, opened 2016-09-15
Add a BitBar plugin to control Logitech Media Server.

michaelherger:
- check whether it's running, using different icons depending on state
- allow to start/stop it
- add links to web UI and PrefPane


tresni:
Remove the abouturl metadata since you aren't pointing it anywhere valid. Otherwise looks good for merging!

PR #579, opened 2016-10-04
update timezone plus

aaronedell:
to shorten city names


matryer:
Please fix conflicting files.

aaronedell:
don't know what that means

matryer:
fixed in https://github.com/matryer/bitbar-plugins/pull/584

PR #580, opened 2016-10-07
Add support for team ids in pagerduty plugin.

aequitas:
Allow selection teams from which to show incidents.


aequitas:
FAO: (@frese)

PR #590, opened 2016-10-21
Add samplerate display plugin

anthonylauzon:
This PR adds a samplerate display plugin.   It's thorough in that it allows you to select from all available audio devices.


tresni:
The linked screenshot does not appear to actually be a screenshot (or maybe the image was corrupted?)  Please update to either remove the image link or use an actual screenshot of your plug.  Thanks!

matryer:
@anthonylauzon can you provide another image?

anthonylauzon:
The image isn't very exciting, but this plugin is very useful for studio engineers and musicians.  the plugin just displays a number, the current sample rate for a particular audio device used by the system.  I'm just removing the image.

PR #595, opened 2016-10-27
changing status icons to use Skycons by DarkSky

arielallon:
see http://darkskyapp.github.io/skycons/


seripap:
Great, i've started work on this in another branch i'm glad someone else got to it!

Does this work for both normal and dark mode for mac?

arielallon:
Yep, i've tested it in dark mode too. :)
bitbar+OSX does a nice job of this if you provide the images in black+alpha only.

i'm still testing it a bit to make sure all the icons work. probably worth giving it a spin too.

arielallon:
some of the icons look a little small, so i may see what i can do to re-size them. i had them at 22px^2 at 144 dpi.
it's possible i'm following the wrong guidelines, though.

tresni:
@arielallon do you want me to hold of merging this?

arielallon:
yep, the icons are really too small to see.
haven't had a chance to fix that up yet. may be able to shortly and i'll ping you when they are.

arielallon:
ok fixed it up and fully tested (at least on my machine)
from my perspective, this is good to go.

arielallon:
woops, forgot to explicitly ping you in the above message.
@tresni this is good to go.

Issue #525, opened 2016-08-17
Move package_manager.py to its own repository


kdeldycke:
The [`package_manager.py`](https://getbitbar.com/plugins/Dev/package_manager.7h.py) script I created some weeks ago saw tremendous interests by the community. See: #466, #469, #479, #480, #482, #493, #498, #516, #520 and #523.

With time the script will grow bigger. Simply because we'll try to make its behaviour consistent and predictable in the face of highly variable environments. And also because we'll probably try to reduce the discrepancies between package managers.

We're not far from the symbolic size of 1000 lines, and I fear that any additional feature is going to bring it on the verge of unmaintainability.

Hence my proposal to move it to its own repository. This will allow us to let the script gets dedicated version tags, development and stable branches, documentation, unit-tests, version history, bug tracker, continuous integration and automatic code style checks.

I'm also thinking about putting all management code to a dedicated standalone Python package. This would add a dependency on the script itself, but nothing too complicated. We'll just need to `$ pip install package-manager`. This will greatly simplify the script itself.

What do you guys think of my proposal?

CC: @tresni, @keithamus

keithamus:
> I'm also thinking about putting all management code to a dedicated standalone Python package. 

This sounds great. Make libraries not applications! It'd be great to repurpose this for other things.

tresni:
I totally think it makes sense to move this to it's own repo for tracking changes, releases, etc.  Can still have the file included here too.

That is, unless you split it out into libraries...  If it's a `pip` installed library it should be fine, but just having a dependent module in the same folder won't (see standard response to any plugin that is more then a single file.)

kdeldycke:
OK, so here is what I plan to do:

1. Isolate the script commit history and move it to a dedicated repository (already did that a couple of times).
2. Continue working on it over there, like we did in the past weeks, like nothing happened.
3. Push a copy of the standalone script here every time we release a tagged version.

That's all.

And while we keep working remotely on the standalone script, we'll experiment on a library on the side, in the same repository.

This might temporarily lead to duplicate code, but we'll have the whole project in one centralized place. There we'll have enough space to slowly think and lay out a plan for a painless transition, from the standalone script to a bare bitbar-script + a library.

kdeldycke:
Here you go => https://github.com/kdeldycke/package-manager

tresni:
I'm in :)

kdeldycke:
[...]

PR #625, opened 2016-12-14
Upgrade to Meta Package Manager v2.1.0.

kdeldycke:
This is the first BitBar plugin of the v2.x series, which is now entirely dependent on the [`mpm` Python CLI](https://pypi.python.org/pypi/meta-package-manager).

Relying on `mpm` makes the BitBar plugin tiny and easier to maintain: all the logic is now coded in the external `meta-package-manager` Python module. This include searching for package manager, synchronizing repositories, listing outdated packages, upgrading them and formatting BitBar menus.

It also makes the BitBar plugin future proof, as support for new package manager will be implement in the `mpm` CLI.

This concludes the migration process discussed 4 months ago in #525 .

### DIFF
--- a/Dev/MetaPackageManager/meta_package_manager.7h.py
+++ b/Dev/MetaPackageManager/meta_package_manager.7h.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # <bitbar.title>Meta Package Manager</bitbar.title>
-# <bitbar.version>v1.11.0</bitbar.version>
+# <bitbar.version>v2.1.0</bitbar.version>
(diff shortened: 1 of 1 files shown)


kdeldycke:
Oh, and to migrate from the previous v1.x series to the new v2.x, just upgrade it as you do normally for any BitBar plugins. On the first run, the new version will feature a new menu entry labelled `Install / upgrade Meta Package Manager.` if it can't find the new `mpm` dependency:




Clicking this item will open a terminal window and install the `mpm` CLI on the system. Then a final refresh of the plugin should show you the usual menu.

kdeldycke:
Thanks @tresni for the merge! 👍

Issue #624, opened 2016-12-14
Remaining time on battery replacement


iosdeveloper:
Need a plugin for that

tresni:
https://github.com/holman/dotfiles/blob/master/bin/battery-status

PR #629, opened 2016-12-18
Add Battery percentage and time remaining plugin

eripa:
Show current Battery change percent and time remaining (dropdown)

battery-status


tresni:
Fixes #624

PR #634, opened 2016-12-25
Upgrade Meta Package Manager plugin to v2.2.0.

kdeldycke:
This version adds a new alternative table-like rendering, with all packages placed in submenus, one per package manager. This alternative version is not active by default.


kdeldycke:
Thanks @tresni for the merge! :)

PR #648, opened 2017-01-09
add gitlab status

blockloop:
image


tresni:
See [the errors](https://travis-ci.org/matryer/bitbar-plugins/builds/190440974#L1090-L1102).  Specifically it looks like you are using `/bin/sh` but the conventions you are using are not available in `sh`.  May want to change your shebang to use `/bin/bash` instead.

blockloop:
Ah yes. That's such a terrible habit. Pushed a change to use bash.

PR #649, opened 2017-01-10
Hosting plugin screenshot at github instead of on main site to avoid remote linking problem

monerohow:
Image was showing as broken before this fix due to Cloudflare remote linking issue


tresni:
We don't allow images in this repo.  just the scripts themselves at this time.

PR #662, opened 2017-01-17
Copy to clipboard Plugin

parvez:


tresni:
Fixed line endings in https://github.com/matryer/bitbar-plugins/commit/6a53b4535f2b2258bf033f8358e6cb6f05db4f44

PR #669, opened 2017-01-23
New iTunes Lite Plugin

prenagha:
Wrote a new BitBar plugin that reports on iTunes status. Slightly more lighweight than the others. No images and such. And runs as compiled applescript so hopefully fast. Please add to plugin repo.
More info at
https://github.com/prenagha/bitbar-itunes


prenagha:
will fix per shellcheck and resubmit PR

PR #670, opened 2017-01-23
new plugin: iTunes Lite

prenagha:
Wrote a new BitBar plugin that reports on iTunes status. Slightly more lighweight than the others. No images and such. And runs as compiled applescript so hopefully fast. Please add to plugin repo.
More info at
https://github.com/prenagha/bitbar-itunes


prenagha:
Not noticing the plugin on getbitbar.com
Is there something else I need to do to get it to show up there?

tresni:
It only rebuilds once a day.  It's there now :) https://getbitbar.com/plugins/Music/bbitunes.10s.sh

PR #673, opened 2017-01-24
Plugin to delete temporary files from Visual Studio  Code

AlexPoulsen:
Visual Studio Code leaves a lot of temporary files around after running code and creates new ones anytime code is run, even if the old temp files are still there and the code is untouched. This deletes all files in the folder it is pointed to that start with 'temp'.

### DIFF
--- a/System/deleteVSCodeTemp.sh
+++ b/System/deleteVSCodeTemp.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# <bitbar.title>Delete VSCode Temp Files</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author.github>AlexPoulsen</bitbar.author.github>
+# <bitbar.author>AlexPoulsen</bitbar.author>
+# <bitbar.desc>Deletes Visual Studio Code Temporary Files</bitbar.desc>
+
+if [ "$1" = 'clean' ]; then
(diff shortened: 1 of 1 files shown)


AlexPoulsen:
-r breaks deletion check

tresni:
I'd suggest reading [SC2162](https://github.com/koalaman/shellcheck/wiki/SC2162) on shellcheck's wiki, but if you feel that without the `-r` is the correct way, then you can override the check.    On the line preceeding the `read` put:

```sh
# shellcheck disable=SC2162
```

You do need to `chmod +x` your script.

tresni:
Side note: You don't need to close/reopen the PR.  If you update your remote repo, the PR will automatically update with your changes.

AlexPoulsen:
I had run chmod +x on it, but I'll try that again.
Thanks for the side note

AlexPoulsen:
The offending backslash isn't before a space or a line feed and I don't think it is necessary.

AlexPoulsen:
Am I doing something wrong? I ran chmod +x (link to file) several times and it still isn't executable.

tresni:
@AlexPoulsen Should just be:

```sh
chmod +x System/deleteVSCodeTemp.sh
git add System/deleteVSCodeTemp.sh
git commit -m "chmod +x'd the thing"
git push origin master # assuming your upstream is named origin
```

PR #676, opened 2017-01-25
Create todotxt.15s.sh

denzuko:
Added plugin for todo.sh intergration


tresni:
@denzuko Please review the [output of the TravisCI tests](https://travis-ci.org/matryer/bitbar-plugins/builds/195036102).  Looks like you need to `chmod +x` your script and wrap the subshell call in quotes if you want it to be passed as a single option

PR #677, opened 2017-01-25
New Plugin:Instagram stats

JanGross:
in web/Instagram/igs.10m.py


tresni:
Looks like you import `random`, `sys`, and `from PIL import ImageDraw` but don't use them.

JanGross:
@tresni 
I didn't notice those leftover imports. 
Removed them in 7c4a542

PR #691, opened 2017-02-08
fix misleading description

drselump14:


tresni:
@drselump14 you'll need to update to a direct link to the image. If you change `dl=0` to `raw=1` I think it should work.

drselump14:
@tresni ok, you got it

PR #696, opened 2017-02-12
Add checkhosts

christophschlosser:
I still can not get this to work in the interpreted version.

See https://github.com/matryer/bitbar/issues/394 for more details


tresni:
Until this has a known way to work I'm going to hold off merging it.  We'll also need to find a linter for golang to add to the travis-ci build

PR #697, opened 2017-02-13
Update plugin: Dev/Vagrant/vagrant.30s.pl

artifactsauce:
I will propose revising the plugin which I also use conveniently.
The plugin was originally developed by @alexandregz .

The revise is enable to copy the path to the vagrant directory to the clipboard.

---
This change is [Reviewable](https://reviewable.io/reviews/matryer/bitbar-plugins/697)


artifactsauce:
Thank you for pointing out 😄  Fixed it ✏️

PR #707, opened 2017-02-27
Add a plugin to display your time on earth

timlapluie:
This adds a plugin to display how long you are already living on this planet:

screen shot 2017-02-27 at 5 53 36 pm

This was created to not only have a plugin for the remaining time. Sometimes it's good to think positive 😉


tresni:
I like it!

PR #711, opened 2017-03-02
Create Heroku List Apps plugin

cody1213:
This is a simple plugin that uses the Heroku toolbelt to list all the apps on your account and provide links to both the app and the app's Heroku dashboard.


tresni:
You need a file extension. In this case use `.sh` to indicate that it's a shell script.

PR #727, opened 2017-03-22
New plugin showing current moon phase as emoji

hpcodecraft:
A plugin that shows the current moon phase.

Moon phase plugin screenshot


matryer:
Lovely

PR #738, opened 2017-04-05
Fixed https

aaronedell:
For some reason this stopped working. Adding http’s’ and changing
version to 1.0 fixed it


tresni:
You have a conflict because of the Time/timezoneplus plugin that is included in this PR.

tresni:
Looks like you accidently changed the image link.  Needs to be the i.imgur.com link.

Issue #742, opened 2017-04-08
Re: Habitica | Invalid Array Length


awitherow:
<img width="658" alt="screen shot 2017-04-08 at 16 37 55" src="https://cloud.githubusercontent.com/assets/1827147/24829847/e60eb98c-1c79-11e7-8c0a-c5c9a3cbacdb.png">

First of all, awesome thing! Really excited to use this 👍 

I added my api user and token and I am getting this error.

The code erroring is in this block is here:

```javascript
const progressBar = function(n, total, charLength) {
  n = Math.max(n, 0); // Health can be negative

  charLength = charLength || 10;

  const progress = (n / total) * charLength;
  const ish = progress % 1;

  const filled = Array(Math.floor(progress) + 1).join(FILLED);
  const middle = ish ? FILLEDISH[Math.ceil(FILLEDISH.length * ish) - 1] : false;
  const unfilled = Array(charLength - Math.floor(progress) + (middle ? 0 : 1)).join(UNFILLED);

  return filled + (middle ? middle : '') + unfilled;
};
```

I was wondering if I could get some ideas on how to debug this myself and help!

Thanks.

awitherow:
I managed to get it working by changing the default assignment of charlength:

```javascript
const progressBar = function(n, total, charLength = 10) {
```

matryer:
Please submit a PR to fix the plugin?

PR #752, opened 2017-04-21
refactor(habitica): use of default property fixes undefined

awitherow:
closes https://github.com/matryer/bitbar-plugins/issues/742


tresni:
Pretty sure these should be equivalent but if it works, I'm not going to second guess it :)

awitherow:
Yeah it was from what I can see the same thing... might have been my computer being weird, but I restarted it twice and was not working and then tried this on my computer.

Could have been a local node version something?

Is there a way to require the node engine running behind these menu item apps?

PR #528, opened 2016-08-18
Added currentFiles.1h.rb plugin for editing files

raguay:
I added my plugin to the repository.

PR #768, opened 2017-05-17
TodoNotePlan Plugin

raguay:
Hi,

I finally finished and tested the NotePlan plugin for Bitbar. Could you merge it in? Thanks.

Richard


tresni:
I'd suggest rebasing on upstream/master next time as you are actually overwriting your CurrentFiles plugin that was submitted previous (#528) *and* submitting a new plugin.  I'll accept this time as it doesn't look like there are any changes to the currentFiles plugin in here.

PR #779, opened 2017-05-31
Create CryptoPrices.5m.sh

viiraj:
A plugin that rotates the prices of BTC-USD and BTC-9 other Cryptocurrencies


tresni:
`chmod +x` your script, commit, and push and it should be good to go!

viiraj:
Thanks @tresni - I have changed the permissions and created a PR. I tried to push but not sure I have permissions?

tresni:
You've added an `ß` in your shebang (`#!`) which breaks things :(

PR #781, opened 2017-06-02
Handle Taskwarrior task list alignment issues with multi-digit IDs

### LINKED ISSUE #780: re: Taskwarrior
FAO: (@christophruss)

Columns of tasks with multi-digit IDs are misaligned due to the string display being stripped.

Illustration attached.

alignment issues

I'll propose a PR soon.

agravier:
Places square brackets around IDs, preserving right padding. A rstrip alone is not enough, as the padding gets mangled later within BitBar, before display.

Resolves  #780

Tested to work with an empty task list and with 1 and 2 digits task IDs

alignment fix

### DIFF
--- a/Lifestyle/taskwarrior.4m.py
+++ b/Lifestyle/taskwarrior.4m.py
@@ -115,11 +115,12 @@ def print_output(
         print '---'
 
     for content_line in content_lines[2:-1]:
-        content_id = content_line.split()[0]
-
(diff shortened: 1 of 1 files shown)


agravier:
This PR breaks commands. I'll fix this, please don't merge. Thanks.

agravier:
Fixed. All commands and alt commands work now.

PR #787, opened 2017-06-10
Pubmed publications

ozturan:
**Pubmed**
Gives the number of publications for a given keyword on [Pubmed](https://www.ncbi.nlm.nih.gov/pubmed). Just returns the search result for that keyword. It is useful to take a peek from time to time. 
PubmedSS


tresni:
You appear to have made an empty change.  you added your code and then deleted it in the last commit.

ozturan:
Fixed.

PR #788, opened 2017-06-12
Coinbase plugin to show BUY prices for multiple coins

dkvasnicka:


tresni:
So just using the compiler (with optional rackjure extension) as the linter?  No other specific linter?

dkvasnicka:
Yep, the Racket world doesn't really use any specific standalone linter to check the code. `make` obviously doesn't check the formatting and code style conventions but other than that it is pretty thorough, definitely so for a dynamic programming language. (e.g. https://github.com/mcandre/linters only mentions Typed Racket as a bonus, nothing else)

PR #792, opened 2017-06-20
Pull Request: sbcdash_client

phntxx:
Hello,
I created a little plugin for BitBar that interacts with an API that I wrote.
Basically, it just pulls the CPU temperature, RAM and drive usage from the API running on a single-board-computer (like Raspberry Pi, Pine64 or BananaPi) and puts it out in BitBar.
There are no dependencies other than Python, so it should run just fine.
If you need an example output file (like a sample of what the API would provide) it can be found under
https://phntxx.github.io/testing/sbcdash.json

I would really appreciate it if you could add this plugin to the repository.

- phntxx


tresni:
Your script needs to be made executable.  This can normally be accomplished by doing the following:

```
chmod +x /path/to/your/script.sh
git add /path/to/yout/script.sh
git commit -m "Making script executable"
git push origin master
```

phntxx:
Just completed the steps you mentioned above - do I need to file a new pull request or has this issue been resolved just by pushing a new version into my fork?

tresni:
Don't need to open a new PR or anything.  Travis-CI will re-test, that was the only issue I saw so as soon as Travis CI says it's good, then we'll get this merged.

PR #793, opened 2017-06-22
Update countdown.1s.py

Furkanzmc:
Make changes to accept arguments for dynamic times and labels.
The first none date argument will be treated as a label and the one after that will be accepted as date.

Available Arguments:
    --bar-title: This will appear as the first line in the output. The default is 'Countdown Timer'.
    --date-format: You can provide a custom date format. The default is '%d-%m-%Y %H:%M'
    --no-cycle: If this is present in the arguments, the times will not cycle.
    --help: Prints this message and exits.

Screenshot
image

Example Usage:

    countdown.py --bar-title "Custom Bar Title" --no-cycle "--date-format" "%d-%m-%Y" "Time #1" "17-07-2017" "Time #2" "15-08-2017"

### DIFF
--- a/Time/countdown.1s.py
+++ b/Time/countdown.1s.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 # <bitbar.title>Countdown</bitbar.title>
-# <bitbar.version>v1.0</bitbar.version>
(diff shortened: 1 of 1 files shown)


tresni:
Due to the extensive changes, would you please increase the version number in the plugin metadata?

PR #794, opened 2017-06-23
Pagerduty API v2

frese:
Upgraded pagerduty.s30.rb to use API v. 2


tresni:
@frese Thanks for the PR, could you please increment the version # in the metadata?

frese:
But of cause I should :-)

PR #795, opened 2017-06-24
adding my crypto trackers

theocean154:
Three scripts, pulled from the works of many others. figured I'd give them back.


tresni:
we require attribution on all scripts.  So you need to add `<bitbar.author>` and `<bitbar.author.github>` metadata to the scripts before we can accept them.

PR #823, opened 2017-08-04
Display Coinbase Prices

mustafaturksavas:
Displays Bitcoin, Ethereum and Litecoin prices from Coinbase

### DIFF
--- a/Cryptocurrency/CoinbasePrices.sh
+++ b/Cryptocurrency/CoinbasePrices.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# <bitbar.title>Coinbase Prices</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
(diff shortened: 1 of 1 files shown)


mustafaturksavas:
Hi, I've tried this for the second time, but getting the same problem. In both instances, I've run `chmod +x plugin.sh` command before uploading. May this be because I'm using github's file upload feature?

marblenix:
@mustafaturksavas,

You are correct. Github's file upload does not respect the executable bit on files.

There's also a couple other mistakes I found, such as 2017-04-16 not being a valid CB-VERSION for coinbase. You also have to be careful when parsing JSON data inside bash. JSON does not guarantee the order in the response, even though it doesn't change for most API implementations. It's better to use a JSON parser such as [jq](https://stedolan.github.io/jq/).

I made a pull request for you here: https://github.com/mustafaturksavas/bitbar-plugins/pull/1

PR #828, opened 2017-08-08
Update Umbrella Status to v1.1, add user-friendly password prompt

jasmas:


jasmas:
Hold off. Will fix lint check complaints and add image metadata.

PR #833, opened 2017-08-11
Update yahoo-stock.1m.sh

foice:
for me it works only if I use 
```
`echo "${s/stock_symbol/${stock[$c]}}"`
```

### DIFF
--- a/Finance/yahoo-stock.1m.sh
+++ b/Finance/yahoo-stock.1m.sh
@@ -37,7 +37,7 @@ n=$((n-1))
 
 for (( c=0; c<=n; c++ ))
 do
-	echo -n ${stock[$c]}; echo -n ":"; curl -s echo "${s/stock_symbol/${stock[$c]}}"
+	echo -n ${stock[$c]}; echo -n ":"; curl -s `echo "${s/stock_symbol/${stock[$c]}}"`
 
    # experimental: also log to disk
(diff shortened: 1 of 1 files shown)


tresni:
Please correct the following linter issues and we'll get this merged!

```
In /Users/travis/build/matryer/bitbar-plugins/Finance/yahoo-stock.1m.sh line 40:
	echo -n ${stock[$c]}; echo -n ":"; curl -s `echo "${s/stock_symbol/${stock[$c]}}"`
                                                   ^-- SC2046: Quote this to prevent word splitting.
                                                   ^-- SC2116: Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.
```

foice:
```
curl -s "$(echo "${s/stock_symbol/${stock[$c]}}")"
```
works for me, which is doing SC2046, SC2006 you suggested. However if get rid of `echo` it does not work anymore, because in that way it ends up trying to execute a command. Maybe I am misusing the whole thing, but as far as I knew/know and can see `echo` is essential to return the result of the command.

tresni:
Wow, sorry I missed this.. You can put the following right before line 40:

```sh
#  shellcheck disable=SC2116
```

That will suppress that particular check for that specific line.

dracos:
Sorry to butt in, but you don't need the echo; this works fine and is simpler:
`echo -n ${stock[$c]}; echo -n ":"; curl -s "${s/stock_symbol/${stock[$c]}}"`

PR #837, opened 2017-08-14
Add "People In Space" plugin

mullwar:
How many people are in Space right now?


tresni:
Totally possible to use ES6 if that's what you are comfortable with.  Just add comment right after the shebang that says: `/* jshint esversion:6 */`.  There are a couple other issues that are being caught you'll want to take a look at!

Issue #842, opened 2017-09-06
gfinance.5m.py stopped working


j-gao:
It seems Google's finance api stopped responding.
e.g. `http://finance.google.com/finance/info?client=ig&q=snap` is returning 404 now.

djsmith42:
Looks like Google finally shut it down. They've been threatening to do this for a while. I'm trying to find a Yahoo-based replacement.

Reference: http://googlecode.blogspot.com.br/2011/05/spring-cleaning-for-some-of-our-apis.html

tresni:
We recommend migrating to the Yahoo Finance tracker as the Google Finance API is deprecated.

PR #844, opened 2017-09-07
Migrate to Yahoo finance API since Google is dead now

djsmith42:
Fixes: https://github.com/matryer/bitbar-plugins/issues/842

### DIFF
--- a/Finance/gfinance.5m.py
+++ b/Finance/gfinance.5m.py
@@ -11,20 +11,20 @@ import urllib2
 
 stocks={"MSFT","AAPL","GOOGL","AMZN","ONDK"}
 
-
 stocks = sorted(stocks)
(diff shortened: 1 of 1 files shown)


tresni:
There are already 2 Yahoo Finance stock trackers:

https://github.com/matryer/bitbar-plugins/blob/master/Finance/yahoo-stock.1m.py
https://github.com/matryer/bitbar-plugins/blob/master/Finance/yahoo-stock.1m.sh

Not sure it makes sense to have a 3rd one that is called Google Finance but actually uses Yahoo.  I'd recommend we remove the deprecated script and leave the Yahoo ones for people to use.

djsmith42:
Haha, I didn't realize the "g" in "gfinance.5m.py" referred to Google. Facepalm.

djsmith42:
Since Google is no longer supporting the stock ticker API, this plugin should probably be removed.

PR #856, opened 2017-09-21
Add WaniKani plugin

jakeoid:
For the service (for learning Japanese Kanji) WaniKani, displays some basic information.

Source: https://github.com/jakeoid/bitbar-wanikani


jakeoid:
Script uses f-strings which are only implemented in Python 3.6, perhaps Travis needs redoing?

jakeoid:
Just another thing, generic install location on OSX is `#!/usr/local/bin/python3.6`, or it was when I installed Python3.6!

tresni:
Yeah, for version specific that is correct.  If you only have 3.6 installed then (normally) /usr/local/bin/python3 is symlinked to /usr/local/bin/python3.6 :)

PR #868, opened 2017-10-03
switch to cryptomate

davidosomething:
switch to the cryptomate api since coinmarketcap is down
leave original (refactored) in comments in case someone wants to use it


davidosomething:
oops

PR #874, opened 2017-10-05
Added TOTP OAuth Token Generator plugin

gunasekar:
Added TOTP OAuth Token Generator plugin. This requires [Oauth Toolkit](http://www.nongnu.org/oath-toolkit/) which can be brew installed
`brew install oath-toolkit`


gunasekar:
Fixed the shellcheck errors and resubmit.

PR #875, opened 2017-10-05
Added TOTP OAuth Token Generator plugin

gunasekar:
Added TOTP OAuth Token Generator plugin using [OATH Toolkit](http://www.nongnu.org/oath-toolkit/). OATH Toolkit can be installed using `brew install oath-toolkit`

### DIFF
--- a/Dev/TOTPOpenAuth/totp.20s.sh
+++ b/Dev/TOTPOpenAuth/totp.20s.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# <bitbar.title>Authenticator</bitbar.title>
+# <bitbar.version>v0.1</bitbar.version>
+# <bitbar.author>Gunasekaran Namachivayam</bitbar.author>
+# <bitbar.author.github>gunasekar</bitbar.author.github>
+# <bitbar.desc>This plugin will generate the TOTP tokens and allows to copy them to clipboard</bitbar.desc>
+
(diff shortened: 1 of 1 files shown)


Dev/TOTPOpenAuth/totp.20s.sh
```diff
@@ -0,0 +1,44 @@
...
+# Value - base32 secret key corresponding to the TOTP Account
+vpnsecrets=( "OpenVPN:2b2drladcdoxtpheuom6t4zjsr6tq7ix"
+        "BitBucket:2b2drladcdoxtpheuom6t4zjsr6tq7ix"
+        "GitHub:2b2drladcdoxtpheuom6t4zjsr6tq7ix"
+        "Okta:2b2drladcdoxtpheuom6t4zjsr6tq7ix"
+        "GrabTaxi:2b2drladcdoxtpheuom6t4zjsr6tq7ix" )
+
+# oath-toolkit needs to be installed. Use 'brew install oath-toolkit'
+# update the appropriate path of oathtool binary below
+oathtool="/usr/local/Cellar/oath-toolkit/2.6.2/bin/oathtool"
```
tresni: Would it be better to do `/usr/local/bin/oathtool` so it's more likely to work on newer versions of oathtool?

gunasekar: Made the appropriate change, tested and added to the pull request.

PR #877, opened 2017-10-10
btcturk new api fix.

erhan:
btcturk new api fix.


tresni:
For future reference: with changes like this we generally like to see a version incrementation as well.

PR #878, opened 2017-10-10
fix #769: ping works when when sub 1ms

### LINKED ISSUE #769: Ping plugin incorrectly reports sites as down
If the ping time is <1ms, the plugin will report the site is down.

blockloop:
screen shot 2017-10-10 at 5 02 10 pm

### DIFF
--- a/Network/ping.10s.sh
+++ b/Network/ping.10s.sh
@@ -36,10 +36,12 @@ PING_TIMES=
 
 while [ $SITE_INDEX -lt ${#SITES[@]} ]; do
     NEXT_SITE="${SITES[$SITE_INDEX]}"
-    NEXT_PING_TIME=$(ping -c 2 -n -q "$NEXT_SITE" 2>/dev/null | awk -F '/' 'END {printf "%.0f\n", $5}')
-    if [ "$NEXT_PING_TIME" -eq 0 ]; then
+    if RES=$(ping -c 2 -n -q "$NEXT_SITE" 2>/dev/null); then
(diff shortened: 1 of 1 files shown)


Network/ping.10s.sh
```diff
@@ -36,10 +36,13 @@ PING_TIMES=
 
 while [ $SITE_INDEX -lt ${#SITES[@]} ]; do
     NEXT_SITE="${SITES[$SITE_INDEX]}"
-    NEXT_PING_TIME=$(ping -c 2 -n -q "$NEXT_SITE" 2>/dev/null | awk -F '/' 'END {printf "%.0f\n", $5}')
-    if [ "$NEXT_PING_TIME" -eq 0 ]; then
+    RES=$(ping -c 2 -n -q "$NEXT_SITE" 2>/dev/null)
+    if [ $? -eq 0 ]; then
```
tresni: You are getting a shellcheck error on this line:

```
In /Users/travis/build/matryer/bitbar-plugins/Network/ping.10s.sh line 40:
    if [ $? -eq 0 ]; then
         ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
```

The "correct" way to do this is

```sh
if RES=$(ping -c 2 -n -q "$NEXT_SITE" 2>/dev/null); then
```

That both captures your output and checks that the subshell exits without an error.

blockloop: Fixed. Shellcheck was disabled locally for some reason

blockloop:
Okay submitted a fix

PR #879, opened 2017-10-11
Added Cloudflare Development Mode

aaroncrawford:
Plugin created to quickly manage the development mode of a Cloudflare account to avoid having to login to their web interface all day long as changes are made on your various Cloudflare-based sites.


tresni:
For future reference you could have added:

```js
/* jshint asi: true */
```

This would suppress the [semicolon warnings](http://jshint.com/docs/options/#asi).

aaroncrawford:
Will do - sorry stole that off a boilerplate and just threw in the semis.  Let me know if you need anything else!

PR #880, opened 2017-10-12
Add emojiWeather

Chandler-Davidson:
Publishing a new plugin for BitBar.

### DIFF
--- a/Weather/emojiWeather.30m.js
+++ b/Weather/emojiWeather.30m.js
@@ -15,7 +15,7 @@ let request = require('request');
 
 let apiKey = '';    // https://home.openweathermap.org/api_keys
 let city = 'Huntsville';
-let url = `http://api.openweathermap.org/data/2.5/weather?q=${city}&units=imperial&appid=${apiKey}`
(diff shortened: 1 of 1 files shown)


tresni:
Hi @Chandler-Davidson and thanks for the Pull Request!  We use [jshint](https://www.jshint.com) to lint all Javascript plugins.  It's barfing because you are using some ES6 features (specifically `let`) without saying it should be linted as ES6.  Add this line right under your shebang and it should be good to go:

```js
/* jshint esversion: 6 */
```

Weather/emojiWeather.30m.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/local/bin/node
```
tresni: Add jshint configuraiton here to say `esversion: 6`

Chandler-Davidson: Just committed the new line. Do I need to issue a new pull request or is it set? Thanks for your help!

tresni: Nope, should be automagicly picked up here :)

tresni:
Last error is that your script isn't executable (`chmod +x`)  I'll get this one.  Thanks for the Pull Request!

PR #886, opened 2017-10-15
Add a new plugin: Ethereum price in whole USD

CLL80:
I've seen for a few weeks now that the api used for some of the ethereum price plugins is down (coinmarketcap-nexuist.rhcloud.com). This new plugin uses a working api and also has some other differences from the other ethereum price plugins—it rounds the value to the nearest full dollar and simplifies the display by not showing a ticker symbol.

I've also updated the screenshot URL for the spot-price-buy-sell plugin. The intent is so I can reuse this same repo for other plugin screenshots.


tresni:
Is this any different then https://github.com/matryer/bitbar-plugins/blob/master/Cryptocurrency/Ethereum/eth-usd.py ?

CLL80:
Yes. It omits the decimals, ticker symbol, and uses a different API .

tresni:
The previous API was used (I believe) as no official coinbase API existed at the time.  I would prefer we update those scripts to use the official API instead of having one using the official API, and two not.  Additionally this would be the 3rd script that does the same thing (see [ethusd.1m.sh](https://github.com/matryer/bitbar-plugins/blob/master/Cryptocurrency/Ethereum/ethusd.1m.sh) and [eth-usd.py](https://github.com/matryer/bitbar-plugins/blob/master/Cryptocurrency/Ethereum/eth-usd.py) )

PR #902, opened 2017-11-06
Add a new plugin in Time: Pomodoro Timer

mkoura:
Pomodoro timer with more features than the one that's already included in the repo.
It supports long breaks, can be paused, run script, etc.

It works on Linux with https://github.com/p-e-w/argos and that's where I use it and where I tested it.
I don't have Mac OS X available so I would appreciate if someone could test it there and send me screenshots that I can include in the script's metadata.

I tried to make the script fast and light on resources. It doesn't spawn external processes when possible (like cat, cut, etc.). Apart from notifications and ``date`` on MacOS X it's pure bash. As a result the number of CPU wakeups is very low. It can be run e.g. every 30s and then the impact on resources is really negligible.

### DIFF
--- a/Time/pomodoro2.30s.sh
+++ b/Time/pomodoro2.30s.sh
@@ -1,0 +1,314 @@
+#!/bin/bash
+
+# <bitbar.title>Pomodoro Timer</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Martin Kourim</bitbar.author>
+# <bitbar.author.github>mkoura</bitbar.author.github>
+# <bitbar.desc>Timer that uses Pomodoro timeboxing</bitbar.desc>
+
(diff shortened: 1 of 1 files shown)


tresni:
Hi @mkoura and thanks for the PR!  Before I merge, just wanted to see if there was a reason to make this a separate plugin vs updating the existing one.  Generally (though not always) we try to limit to 1 plugin doing 1 thing, and use that as examples for people to build off.  But we don't hold that hard and fast (see CryptoCurrencies...)

mkoura:
@tresni I have no problem with making this an update of the existing plugin. The reason I created separate plugin is that it has more features and looks different so it can potentially break work-flow of the users of the original plugin.
Also I don't think it's necessary to run it every second, so the file name would differ from the original plugin.

PR #904, opened 2017-11-09
Vertcoin ticker plugins

cemyld:
Adds two cryptocurrency plugins that use bittrex api to show vertcoin prices.


Cryptocurrency/Vertcoin/vtc-btc-bittrex_detail.30s.sh
```diff
@@ -0,0 +1,26 @@
...
+RESULT=$(curl -s "https://bittrex.com/api/v1.1/public/getmarketsummary?market=btc-vtc")
+vtcLast=$(echo $RESULT | egrep -o '"Last":\d+\.?\d{0,5}' | sed 's/"Last"://')
```
tresni:
```
vtcLast=$(echo $RESULT | egrep -o '"Last":\d+\.?\d{0,5}' | sed 's/"Last"://')
```

cemyld:
Should be fixed now!

PR #906, opened 2017-11-11
Add a new plugin countdown_timer_2

fferri:
Improved countdown_timer with UI to set task/time, and better parsing/rendering of time.

### DIFF
--- a/Time/countdown_timer_2.1s.py
+++ b/Time/countdown_timer_2.1s.py
@@ -6,7 +6,7 @@
 # <bitbar.author.github>fferri</bitbar.author.github>
 # <bitbar.desc>Simple countdown timer.</bitbar.desc>
 # <bitbar.dependencies>python</bitbar.dependencies>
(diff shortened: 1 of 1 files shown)


tresni:
Hi @fferri and thanks for the PR!  We generally try to keep to 1 plugin doing 1 thing as an example for users to build off of.  Is there a reason to make this a wholly new plugin vs enhancing either [countdown_timer.1s.rb](https://github.com/matryer/bitbar-plugins/blob/cadb0d76e26319daa7c3bfbb93f263fbee4a608f/Time/countdown_timer.1s.rb) or [countdown.1s.py](https://github.com/matryer/bitbar-plugins/blob/663e9d8c562ddc4353a56d80f31187f9b32a1ecc/Time/countdown.1s.py)?

fferri:
It is based off countdown_timer.1s.rb, but I ported to Python cause I don't know Ruby very well.
None of those scripts has a UI to input data.

PR #912, opened 2017-11-15
LTC-USD tracker;ETH-USD tracker; Slack Notification

mgjo5899:
- Rewrote LTC tracker
- Wrote Slack unread message tracker


mgjo5899:
Some changes were made to original Slack notification after finding out about the Slack API rate limits:
_"In general we allow applications that integrate with Slack to send no more than one message per second. We allow bursts over that limit for short periods"_

Since they allow bursts over the limit, I tried to use that for channel info api calls after gathering all the channel information from each Slack group.

PR #914, opened 2017-11-15
LTC-USD tracker; ETH-USD tracker; Slack Notification

mgjo5899:
- The original LTC-USD tracker didn't work.
- The original ETH-USD tracker didn't work.
- New Slack notification (indicator whether there's new message or not)


tresni:
Hi @mgjo5899  and thanks for the PR!  For future reference, it's a lot easier on us if you do one plugin per PR, just makes review substantially easier.

mgjo5899:
@tresni Sorry about that! I will definitely do one PR per one plugin in the future!

PR #930, opened 2017-12-03
Updated pi-hole plugin

fmartingr:
- Refactored code
- Now supports python3 along with python2
- New black and white icon with retina and dark mode support
  - The old color icon is still there, configurable via a variable.

I wanted to put the Disable/Enable actions inside this script instead of opening a link, but the `bash=$0` does not work if the plugin is written in python apparently. I also tried invoking the python interpreter with the path to script but the `param` does not support spaces and I can't seem to make it work putting double quotes or replacing the spaces with slashes (`Folder\ with\ spaces`). Any ideas?


tresni:
Put the argument in quotes:

```py
print("bash=\"$0\" param=\"something with spaces\"")
```

Still want me to merge this as is or are the changes you want to make that need that information?

fmartingr:
@tresni You can merge it, if I make changes I can do another PR ;)

PR #932, opened 2017-12-05
Added new plugin for crypto currency price ticker

Elaniobro:


tresni:
Thanks for the PR.  We really try to have 1 example plugin that does one thing and allow other to modify.  There are a number of price tickers for crypto currencies all ready and we are not accepting further at this point.

PR #935, opened 2017-12-05
New Plugin: coinmarketticker-ruby

Soleone:
Yet another cryptocurrency ticker getting rates from coinmarketcap.com.

image

This one tries to be super readable and extendable using Ruby.

[Readme here](https://github.com/Soleone/bitbar-coinmarketticker).


tresni:
Thanks for the PR. We really try to have 1 example plugin that does one thing and allow other to modify. There are a number of price tickers for crypto currencies all ready and we are not accepting further at this point.

PR #941, opened 2017-12-15
Added ISC condition monitor

panzertime:
This plugin gets the current status code from SANS ISC ([more info](https://isc.sans.edu/infocon.html)) and displays the current status in the appropriate color.


tresni:
You have to deal with the shellcheck error.  you can either quote the parameters as instructed or suppress the error with a shellcheck directive: `# shellcheck disable=SC2060`

PR #942, opened 2017-12-18
Added GDAX plugin

CSFlorin:
I added a plugin using the GDAX python API to see current prices for BTC, LTC, and ETH on GDAX as well as their 24-hour change. You can open GDAX from the menu and change which crypto you display using an AppleScript menu. Images for each crypto are included. Since I'm using an API and helper scripts, there is a folder with everything needed in so that users don't have to install anything. All that is needed is to change the directory in the first line of the gdax.5s.py script to include the user's plugin directory.


CSFlorin:
ahh sorry about that. unfortunately, i rely on writing to a config file (which I could switch to a dynamoDB but no point) and having a separate shell script to allow the user to change coins without editing the gdax.5s.py script directly. Not bashing you or anything, but I'm not really in favor of getting rid of usability to conform to the single-file guide you have. Perhaps you can look into allowing some folder structure for more complex plugins. Also, most of the files I have are a Python virtual environment to not rely on the user's python and to preinstall all the needed packages.

PR #937, opened 2017-12-10
Realtime Stock Price Plugin.

bmosincat:

PR #945, opened 2017-12-21
Add Stock Ticker plugin to replace broken Yahoo! one

rkanter:


bp1222:
I’m not sure protocol here, but maybe it’d be ok to also delete the couple in the repo that are dead.

tresni:
#937 was merged and uses the same API, providing the same basic information. Closing this as duplicate.

@bp1222 definitely worthy of a pull request to remove any dead/outdate plugins!

PR #952, opened 2018-01-06
Plugin: Bitcoin.co.id Spot Market Prices

hapr05:
a simple cryptocurrency price ticker for bitcoin.co.id market, prices are in IDR


tresni:
Thanks for the PR @hapr05 . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs. There are a number of price tickers for crypto currencies all ready and we are not accepting further at this point.  If you feel yours adds new functionality that should be shared, please do a pull request updating an existing price ticker with your new functionality.

PR #955, opened 2018-01-12
JIRA Search plugin

prenagha:
New lightweight JIRA search plugin


tresni:
Hi @prenagha looks like you also got an iTunes plugin in the branch.  Can you please update the PR to only contain the JIRA plugin?

PR #957, opened 2018-01-13
Modified ltc-usd and eth-usd

mgjo5899:
- Modified ltc-usd and eth-usd to use public API from WorldCoinIndex.


tresni:
Since you changed the filenames, I can't really tell what has changed as it shows complete file replacement.  Additionally you've pulled in all your previous changes since this doesn't appear to be rebased off master.

mgjo5899:
Sorry about that..!
1. I will change the filenames back to the original ones.  Still, they almost completely changed.
2. I am not sure what you mean by I've pulled in all my previous changes; do you mean I have some of the unnecessary commits (such as Merge branch commits)?

Thank you for your inputs.

PR #978, opened 2018-02-14
Create ottawa-weather-in-menu-bar

Hamburgle:
# You will have to change the URL if you don't live in Ottawa. :-) This code may be a bit sloppy but it kinda works.
# if any of you professional coders want to enhance or fix this, feel free. But I like it.
# Actually the last two numbers in this string are probably deg F and deg C but I wasn't sure so I left them both in.

### DIFF
--- a/Weather/ottawa-weather-in-menu-bar
+++ b/Weather/ottawa-weather-in-menu-bar
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# <bitbar.title>Mouse Location</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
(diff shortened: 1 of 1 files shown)


tresni:
Thanks for the PR @Hamburgle . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs. There are a number of weather display plugins already and we are not accepting further at this point.  If you feel yours adds new functionality that should be shared, please either reply with an explanation or do a pull request updating an existing weather plugin to add your new functionality.

See https://getbitbar.com/plugins/Weather for existing weather plugins.

PR #979, opened 2018-02-15
cal.1d.sh: Style view

mizoR:
FAO: (@busterc)

Display for "today" is broken a little bit at least in my environment.  Mainly this PR improves it.  Could you review and comment about my patch?

Anyway, thanks for your simple and good plugin 📆 

### As-is

as-is 2

### To-be

to-be

### DIFF
--- a/Time/cal.1d.sh
+++ b/Time/cal.1d.sh
@@ -17,6 +17,6 @@ echo "$year"
 echo "---"
 
 echo "$last_year"
-cal "$last_year" | while IFS= read -r i; do echo "--. $i | font=courier"; done
-cal "$year" | while IFS= read -r i; do echo ". $i | font=courier"; done
-cal "$next_year" | while IFS= read -r i; do echo ". $i | font=courier"; done
+cal -h "$last_year" | while IFS= read -r i; do echo "--. $i | font=courier"; done
+cal -h "$year" | while IFS= read -r i; do echo ". $i | font=courier"; done
+cal -h "$next_year" | while IFS= read -r i; do echo ". $i | font=courier"; done
--- a/Time/cal.1d.sh
+++ b/Time/cal.1d.sh
@@ -17,6 +17,6 @@ echo "$year"
 echo "---"
 
 echo "$last_year"
-cal -h "$last_year" | while IFS= read -r i; do echo "--$i | trim=false font=courier"; done
-cal -h "$year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
-cal -h "$next_year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
+cal "$last_year" | while IFS= read -r i; do echo "--$i | trim=false font=courier"; done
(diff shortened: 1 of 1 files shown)


Time/cal.1d.sh
```diff
@@ -17,6 +17,6 @@ echo "$year"
 echo "---"
 
 echo "$last_year"
-cal "$last_year" | while IFS= read -r i; do echo "--. $i | font=courier"; done
-cal "$year" | while IFS= read -r i; do echo ". $i | font=courier"; done
-cal "$next_year" | while IFS= read -r i; do echo ". $i | font=courier"; done
+cal -h "$last_year" | while IFS= read -r i; do echo "--$i | trim=false font=courier"; done
```
mizoR: FYI:
### cal: `-h` options

```sh
$ man cal | grep '\-h'
     -h      Turns off highlighting of today.
```

### bitbar: trim

> trim=.. whether to trim leading/trailing whitespace from the title. true or false (defaults to true)
>
> see: https://github.com/matryer/bitbar#plugin-api

busterc: What version of OS are you using? I'm on Sierra and my `cal` does not have the `-h` option, therefore the changes would cause an error:

```sh
$ man cal | grep '^\s*\-'
     -J      Display Julian Calendar, if combined with the -e option, display
     -e      Display date of easter (for western churches).
     -j      Display Julian days (days one-based, numbered from January 1).
     -m month
     -o      Display date of orthodox easter (Greek and Russian Orthodox
     -p      Print the country codes and switching days from Julian to Grego-
     -s country_code
     -w      Print the number of the week below each week column.
     -y      Display a calendar for the specified year.
```
However, I suppose this would work:
```sh
cal -h "$last_year" 2>/dev/null || cal "$last_year" | while IFS= read -r i; do echo "--$i | trim=false font=courier"; done
cal -h "$year" 2>/dev/null || cal "$year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
cal -h "$next_year" 2>/dev/null || cal "$next_year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
```


Also, I must've overlooked bitbar's `trim` option. You're mod for that is indeed better!

gingerbeardman: I see `cal -h` on High Sierra

> Turns off highlighting of today.

busterc: Good to know High Sierra has it. Unfortunately Sierra and probably earlier versions do not. So, like I mentioned above (slightly modified) I would change the PR to the following:
```
cal "$last_year" 2>/dev/null || cal "$last_year" | while IFS= read -r i; do echo "--$i | trim=false font=courier"; done
cal -h "$year" 2>/dev/null || cal "$year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
cal "$next_year" 2>/dev/null || cal "$next_year" | while IFS= read -r i; do echo "$i | trim=false font=courier"; done
```
Then it's backwards compatible! I've actually been using this ^ modified version locally since you originally submitted the PR :)

mizoR: Updated with your suggestion (though there is different a little bit to works correctly with High Sierra), thanks!

PR #981, opened 2018-02-18
Fix for Bitcoin price not working

viiraj:
As requested by @SeepSoak


tresni:
Merging despite the linter errors as those errors are newly introduced by shellcheck and the changes in this script aren't causing them (they exist in the already checked in version.)  For future reference, if you could clean them up it would be nice ;-)

PR #994, opened 2018-03-20
Modify the original code to display a stock in the menubar.

nadler:
# with '#' will be ignored, and an empty message aborts the commit.
# On branch master
# Your branch is up-to-date with 'origin/master'.
#
# Changes to be committed:
#	modified:   Finance/realtime-stock-tracker.5s.py
#
Modify the original code to display a stock in the menubar.
This required us to remove the href, at least in the menubar output.
This allows the menubar to be clicked to show the rest of the list
without launching the browser.
But I removed href for all values since I don't use the href.

### DIFF
--- a/Finance/realtime-stock-tracker.5s.py
+++ b/Finance/realtime-stock-tracker.5s.py
@@ -11,9 +11,6 @@
 
 import json, urllib2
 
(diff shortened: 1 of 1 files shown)


nadler:
Sorry, created by accident.

PR #1001, opened 2018-04-11
Add phpstorm xdebug status plugin

haringsrob:
The plugin is really simple, it checks if the xdebug default port (9000) is in use. If it is then we set the status to on.


gingerbeardman:
Neat. Thanks for your contribution!

PR #1023, opened 2018-05-12
Add BitMEX XBT ticker plugin

horimislime:
This plugin displays latest XBTUSD ticker price on BitMEX market :)

screenshot

GitHub repository is here:
https://github.com/horimislime/bitbar-plugin-bitmex-ticker


tresni:
Thanks for the PR @horimislime . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs. There are a number of price tickers for crypto currencies all ready and we are not accepting further at this point.  If you feel yours adds new functionality that should be shared, please do a pull request updating an existing price ticker.

PR #1025, opened 2018-05-17
Repos and other configs are now read from ~/.repos.yaml

MarcoCabazal:
Fixed: CRLF issue when used with Ruby 2.4.x

### DIFF
--- a/Dev/Git/get_pull_requests_bitbar.1h.rb
+++ b/Dev/Git/get_pull_requests_bitbar.1h.rb
@@ -1,14 +1,13 @@
 #!/usr/bin/env ruby
-
+# frozen_string_literal: true
 
 # <bitbar.title>PR Counts for Github and Bitbucket</bitbar.title>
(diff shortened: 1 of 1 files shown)


tresni:
Hey @MarcoCabazal  can you increase the bitbar.version to indicate this is a breaking change?

Dev/Git/get_pull_requests_bitbar.1h.rb
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-
+# frozen_string_literal: true
 
 # <bitbar.title>PR Counts for Github and Bitbucket</bitbar.title>
 # <bitbar.version>v1.0</bitbar.version>
```
tresni: @MarcoCabazal  please increment the version to indicate this is a breaking change.

MarcoCabazal: Thanks, will do just that.

PR #1028, opened 2018-05-21
Added plugins

RishabhTayal:
- added app center plugin
- added fastlane plugin


tresni:
Generally we want 1 plugin per pull request.  Accepting as I already merged your fastlane plugin so should be a noop there.

PR #1029, opened 2018-05-25
Create prayer-time.1m.py

kaaninan:
That plugin provides to show iftar times for muslims.


tresni:
Merged with a couple changes:

* Made executable
* Removed unused import
* Got rid of int + str concatenation

Issue #1024, opened 2018-05-15
re: Trash Collector


squashmcdermott:
FAO: (@carlsonorozco) hi is it possible to show size of trash in GB rather than number of items? thanks!

carlsonorozco:
@squashmcdermott Let me do a PR for this soon. Thanks for the suggestion. 👍

PR #1033, opened 2018-05-26
Use trash size instead of count

carlsonorozco:
Fix Issue https://github.com/matryer/bitbar-plugins/issues/1024


gingerbeardman:
What do you think about keeping the count as an `alternate=true`?

https://github.com/matryer/bitbar#writing-plugins

gingerbeardman:
Merged with changes:

`echo "$trash_count items 🗑 | alternate=true"`

PR #1048, opened 2018-06-18
Added emoji flags and goalscorers

dg01d:
Made a few changes based on feedback.

Added Emoji Flags to all countries
Added goalscorers to games where available
Match names in dropdown are links to FIFA's official site.

### DIFF
--- a/Sports/worldcup.1m.php
+++ b/Sports/worldcup.1m.php
@@ -28,18 +28,50 @@ require ".bitbar/vendor/autoload.php";
 
 use SteveEdson\BitBar;
 
+function array_msort($array, $cols)
+{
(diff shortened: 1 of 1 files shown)


tresni:
Generally if you add functionality we like to see a version increment. But given World Cup I’m going to allow it!

gingerbeardman:
Cool plugin.

I think it would be neater if it:
- showed only flags
- maybe with the names revealed as an alternate
- and used three letter ISO Country Codes  

eg.

🇨🇴 0 – 0 🇯🇵
COL 0 – 0 JPN

or at least to have things symmetrical

COL 🇨🇴 0 – 0 🇯🇵 JPN

PR #1050, opened 2018-06-19
World Cup modifications

dg01d:
Took PR from @gingerbeardman 

Needed to add some colour-codes to the script to make that version readable when menu is set to dark mode. If there's a better solution, I'm open to it.


matryer:
@dg01d please resolve conflicts

PR #1052, opened 2018-06-21
Added decode to clipboard plugin

ku16610:
Hi , Just added decoding function to the clipboard base64 encoder plugin...


matryer:
Please make it executable as per `clipboard-base64-encoder.sh not executable` in the CI report.

ku16610:
Hi matryer, so sorry will fix when I’m back near a computer an resubmit.

Kind regards

Richard

ku16610:
Should be fixed now.

Issue #1062, opened 2018-07-04
PagerDuty plugin bad param for API v2


joariasl:
The next line:
https://github.com/matryer/bitbar-plugins/blob/ec3ac184198936271503018330314fa38ad63a1a/Tools/PagerDuty/pagerduty.30s.rb#L75
contains a bad parameter passed to the used API v2.

For query based on teams is required the parameter name `team_ids[]` like array and not using the v1 parameter name.

See: https://v2.developer.pagerduty.com/v2/page/api-reference#!/Incidents/get_incidents
Old API version not used or working in this implementation: https://v1.developer.pagerduty.com/documentation/rest/incidents/list

Replacing `teams` to `team_ids[]` works for one team.

PR #1063, opened 2018-07-04
Fixed matryer/bitbar-plugins#1062

joariasl:
Passing API v2 paramenter correctly


tresni:
Merging as this is a bug fix, but for future updates, you'll need to address the linter errors.

PR #1067, opened 2018-07-11
Update ColorfulBatteryLevel.5s.sh

Phlooo:
Removed name from header


Phlooo:
Why do the check fail since I didn't change any line of code (only the name in header) ?

PR #1068, opened 2018-07-11
Update DarkModeToggle.sh

Phlooo:
Removed name from header


Phlooo:
Why do the check fail since I didn't change any line of code (only the name in header) ?

tresni:
PRs are subject to all new checks that a linter may have (linters improve with time), whereas we attempt to only apply the base linters to existing code (to keep the build from failing every time a linter is updated.)  So yes, you may have only changed the header, but now there more checks from the linter and your plugin is failing.

I'll merge in this case.  But be aware you name will still be there in the git history, just not on the most recent build.

PR #1071, opened 2018-07-16
New Plugin: Hacker News Headlines Lite

amrrs:
Python based and does not display a lot of information (unlike the existing Ruby based Plugin).

### DIFF
--- a/Web/HackerNews/hn_front.120m.py
+++ b/Web/HackerNews/hn_front.120m.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env -S PATH="${PATH}:/usr/local/bin" python3
+
+# <bitbar.title>Hacker News Headlines - Lite</bitbar.title>
+# <bitbar.author>amrrs</bitbar.author>
+# <bitbar.author.github>amrrs</bitbar.author.github>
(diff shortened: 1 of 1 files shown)


Web/HackerNews/hn_front.120m.py
```diff
@@ -0,0 +1,38 @@
...
+print("HN")
+#check if internet connection is available
+#@yasinkuyu 08/09/2017
+#source: https://gist.github.com/yasinkuyu/aa505c1f4bbb4016281d7167b8fa2fc2
+
+def check_internet():
+    url='http://www.google.com/'
+    timeout=5
+    try:
+        _ = requests.get(url, timeout=timeout)
```
tresni: Don't need the assignment here.  Also, expensive to check for internet connection this way, might as well just do the `get()` on line 30 and if you fail there output the "No connection" and `sys.exit` or something.

tresni:
Your script needs to be made executable.  This can normally be accomplished by doing the following:

```
chmod +x /path/to/your/script.sh
git add /path/to/yout/script.sh
git commit -m "Making script executable"
git push origin master
```

amrrs:
Thanks. I'll close this and submit a new PR

PR #1078, opened 2018-07-27
Added plugin that reads news from fcbarcelona.dk

tiktuk:
As the title says.


tresni:
Hi and thanks for the PR!  We use pyflake8 to lint all python plugins. Yours has come back with a number of errors and warnings. Please review the output of the [TravisCI build](https://travis-ci.org/matryer/bitbar-plugins/builds/413686412?utm_source=github_status&utm_medium=notification) and address as appropriate! Once the build comes back clean we can merge the changes!

tiktuk:
Hi @tresni, it should be fixed now :) .

PR #1084, opened 2018-08-24
Add support for lua plugins + network bandwidth plugin in lua

cpbotha:
This plugin shows current total bandwidth over user-specified interfaces, with separate interfaces in context menu.

It is written in lua with a single call to ifstat. It is 30% smaller (peak memory use) and slightly faster than the bash + ifstat + awk versions.


cpbotha:
I see that the Travis CI build failed because: `/Users/travis/build/matryer/bitbar-plugins/Network/bandwidth_total.2s.lua unrecognized file extension`

Would you consider adding lua? It's a single `brew install lua` away. :)

cpbotha:
I have added support for lua plugins, including the luacheck linter.

tresni:
Normally, we don't like to have multiple plugins doing the same thing.  But given you've added support for a whole new language, I'm going to allow it.

cpbotha:
Thank you for merging, much appreciated!

Hopefully now we'll see more small lua plugins.

PR #1088, opened 2018-08-28
[NEW PLUGIN] Sensibo Plugin

niladam:
I've created a plugin that shows my current Sensibo pods, allowing me to also have some basic control over them.

The plugin looks like this:

Sensibo Bitbar Preview

For suggestions and/or issues -- the repo containing this plugin is can is [@niladam/sensibo-bitbar](https://github.com/niladam/sensibo-bitbar).

### DIFF
--- a/Tools/sensibo.60s.sh
+++ b/Tools/sensibo.60s.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+#
+# <bitbar.title>Sensibo PODS</bitbar.title>
+# <bitbar.version>v1.1</bitbar.version>
+# <bitbar.author>Madalin Tache</bitbar.author>
+# <bitbar.author.github>niladam</bitbar.author.github>
+# <bitbar.dependencies>jq</bitbar.dependencies>
+# <bitbar.desc>Provides Sensibo pods (sky v2) status in your menubar allowing you to turn them on/off.</bitbar.desc>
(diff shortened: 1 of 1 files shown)


niladam:
@tresni - this CI failed but not because of the plugin (which failed initially but was fixed). Here's the faillog..

```
==> Downloading https://mirror.racket-lang.org/installers/7.0/racket-7.0-x86_64-
curl: (60) SSL certificate problem: Invalid certificate chain
More details here: https://curl.haxx.se/docs/sslcerts.html
curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
Error: Download failed on Cask 'racket' with message: Download failed: https://mirror.racket-lang.org/installers/7.0/racket-7.0-x86_64-macosx.dmg
The command "brew cask install racket" failed and exited with 1 during .
Your build has been stopped.
```

Can you look into it ?

tresni:
Re-ran the build and it's good to go!

PR #1090, opened 2018-09-02
top 5 memory-consuming applications

hypersport:


gingerbeardman:
These are processes rather than applications. Also, there are numerous different memory usage figures, each measured in a different way, some include shared memory and others do not.

gingerbeardman:
Is the information in the screenshot useful to you? It just shows Safari itself and 4 Safari tabs (but which ones?).

hypersport:
@gingerbeardman How about this one? 
default

PR #1091, opened 2018-09-03
Add strava plugin

danielecook:
Adds a Strava plugin that shows the total lifetime miles run or biked, and the breakdown by the past month and year to date. Requires the user to enter a Strava API access token which is an option available within the dropdown menu.


tresni:
Hi Daniel, we don't accept binary plugins.

PR #1093, opened 2018-09-03
syncthing: fix incorrect cpu usage

blockloop:
CPU usage was incorrectly multiplied * 100


blockloop:
Could use :eyes: here @tresni :)

PR #1104, opened 2018-09-25
Disk Usage: show in a dimmed colour when available space is less than…

sohocoke:
… 2% of capacity.

### DIFF
--- a/System/mdf.1m.sh
+++ b/System/mdf.1m.sh
@@ -99,7 +99,7 @@ make_bmp_header() {
         "$_height"                # height
         01 00                   # 1 color plane
         20 00                   # 32 bits per pixel
-        $comp 00 00 00          # compression
+        "$comp" 00 00 00          # compression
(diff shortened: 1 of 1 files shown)


tresni:
Hi and thanks for the PR!  We use shellcheck to lint all shell plugins. When a plugin is updated, we expect it to meet current best practices.  Yours has come back with a number of errors and warnings. Please review [the output at Travis CI](https://travis-ci.org/matryer/bitbar-plugins/builds/432937376?utm_source=github_status&utm_medium=notification) and address as appropriate! Once the build comes back clean we can merge the changes!

sohocoke:
Hi. The lint error is happening on lines which have not been edited. Perhaps the lint rules have changed since the last changes made on the file. 

The error is easy to address so I'll still fix them and add a commit.

sohocoke:
@tresni took a while, but the lint issues are fixed now.

PR #1116, opened 2018-10-25
Update Web/faith_in_humanity_score.py

impshum:
Added my new website data.


impshum:
Oops. There is more...

PR #1125, opened 2018-11-12
Host switcher improvements

canemacchina:
Added a command to flush OSX dns cache without change the host file


tresni:
For future reference, if you are making a change, we generally like to see a version change (major/minor is up to you.)  That way if people are trying to decide if they want/need to update, they have a quick thing they can check.

PR #1135, opened 2018-11-20
update images and cleanup

frakman1:
New plugin image:
image

### DIFF
--- a/Tools/addnote.5s.py
+++ b/Tools/addnote.5s.py
@@ -4,7 +4,7 @@
 # <bitbar.title>Add Note</bitbar.title>
 # <bitbar.author>Frak Nuaimy</bitbar.author>
 # <bitbar.author.github>frakman1</bitbar.author.github>
-# <bitbar.image>http://i62.tinypic.com/1zvae6o.jpg</bitbar.image>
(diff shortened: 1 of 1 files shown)


gingerbeardman:
I'll leave this for a while, please try to sit on your changes before doing a PR.

gingerbeardman:
The @ denotes a symbolic link.

from `man ls`:

>  -F      Display a slash (`/') immediately after each pathname that is a
>              directory, an asterisk (`*') after each that is executable, an at
>              sign (`@') after each symbolic link, an equals sign (`=') after
>              each socket, a percent sign (`%') after each whiteout, and a ver-
>              tical bar (`|') after each that is a FIFO.

Does that help?

frakman1:
Thank you sir. I looked it up too and I realize that it signifies extended attributes.  What I don't know is *why* it happens to my file after the git operations to check it in. I found instructions on how to remove it but it keeps happening after the git operations.

But more importantly, it highlights a bug in the Travis CI checker because this should not be a disqualifying error as anyone can download and use this file even with the @ extended attribute. Do you know what script/logic/algorithm Travis is using to conclude that the file is not executable? [Here](https://travis-ci.org/matryer/bitbar-plugins/builds/457264608) is the error:

image

PR #1137, opened 2018-11-21
Nightscout plugin

badgerpapa:
This is a plugin specifically for Diabetics who use a Continuous Glucose Monitor and who track their data using Nightscout (http://www.nightscout.info/). It just does a simple REST call and formats the data for BitBar

### DIFF
--- a/Lifestyle/nightscout.1m.sh
+++ b/Lifestyle/nightscout.1m.sh
@@ -29,25 +29,25 @@ fi
 
 case ${RESULTARRAY[3]} in
 	FortyFiveUp)
-		TREND="/"
+		TREND='/'
 		;;
 	FortyFiveDown)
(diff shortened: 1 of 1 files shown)


badgerpapa:
Last check failed with this error: 
 ^-- SC1003: Want to escape a single quote? echo 'This is how it'\''s done'.

I do not want to escape a single quote, I want to display a backslash.

tresni:
Check https://github.com/koalaman/shellcheck/wiki/SC1003 for details. If my understanding is correct, you can either suppress these warnings (as you want a backslash, it just looks like you are trying to escape a single quote) or you can do the backslash outside of the quotes.

badgerpapa:
Using unicode arrows rather than slashes. Prettier and works better.

PR #1139, opened 2018-11-23
New plugin: "sent/received"

rnkn:
Displays total sent and received wifi data for current sesssion.


rnkn:
Thanks. I disabled a check for double quoting because I'm not sure how it would function with double quotes. Seems to be okay?

rnkn:
Sorry you may want to pull the screenshot again from the provided URL... the existing screenshot was counting MBs as GBs.

gingerbeardman:
OK will look into it

PR #1145, opened 2018-12-14
Revert recent commits to fix lint warnings flagged by a disabled rule.

sohocoke:
Fixes regression for mdf.1m.sh where rendering colour is incorrect when disk capacity is low.

### DIFF
--- a/System/mdf.1m.sh
+++ b/System/mdf.1m.sh
@@ -90,16 +90,16 @@ make_bmp_header() {
     # Common bits for version 1 and 5
     bmp_header+=(
         42 4d                   # "BM" magic
-        "$_filebytes"             # size of file
+        $_filebytes             # size of file
         00 00                   # reserved
         00 00                   # reserved
-        "$_pixoffset"             # offset of pixel data
-        "$_headerbytes"           # remaining bytes in header
-        "$_width"                 # width
-        "$_height"                # height
+        $_pixoffset             # offset of pixel data
+        $_headerbytes           # remaining bytes in header
+        $_width                 # width
+        $_height                # height
         01 00                   # 1 color plane
         20 00                   # 32 bits per pixel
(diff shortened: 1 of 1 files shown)


sohocoke:
@tresni a quick question with the shellcheck linting: doesn't the commit below specify that rules SC2206 and SC2207 should be excluded?

https://github.com/matryer/bitbar-plugins/commit/9bebf98953dcb98d4b1dc52887b1b058b60600f7#diff-2f227f602f9e62ae3a54903dcca58585R95

This is a reversion of the changes specifically for those linting rules in #1104, since the script was written with extensive usage of conversions between space-delimited strings and arrays that were non-trivial to re-write just to satisfy linting warnings.

tresni:
Close. We ignore new lint codes for existing code. If a plugin is touched for any reason, or a new plugin is added, we require that it pass the full suite of linter tests.

gingerbeardman:

You can close and reopen the PR to get Travis to try again, see #1214

sohocoke:
@gingerbeardman #1214 appears to be about the build occasionally failing to download deps. This PR is failing CI due to existing code that was committed prior to this PR which now fails lint rules which are getting re-applied. Unless the lint rules have changed, re-running CI would result in the same failure.

I can’t commit to fixing all those lines in the brittle shell script that I didn’t write, and ensuring no regressions, just to pass new lint rules for a plugin for a project which is now hardly maintained — e.g. https://github.com/matryer/bitbar/issues/504 is about the CI for the BitBar project that is broken and is blocking all PR’s since 2017. 

I would suggest changing the lint policy so new PR’s don’t get barred by new lint failures to old code, but since the project doesn’t seem really maintained, it feels like a lost cause.

Closing the PR since it will probably never be merged.

gingerbeardman:
I agree with your thoughts on all of this. Thanks for your contributions.

PR #1152, opened 2018-12-26
Robinhood.1m.py Portfolio Plugin

JonathanGrant:
Just need to install https://github.com/Jamonek/Robinhood with `pip install robinhood`


gingerbeardman:
Thanks!

PR #1155, opened 2018-12-31
Fix an issue in CalendarLite

WeibingChen17:
When the new year is coming, the "next month" should be "Jan 2019" rather than "Jan 2018". 
Fix it now. 

Please pick 7c397ac only.

Happy new year!


WeibingChen17:
Fix #1159

gingerbeardman:
Thanks!

PR #1186, opened 2019-02-21
Simpler copy-to-clipboard.sh

guy4261:
Instead of sometimes passing copy, use bash properly to run pbcopy.


gingerbeardman:
Thanks!

PR #1206, opened 2019-03-26
Adding bitbar-slack-team-notifications

benjifs:
I wrote a variation on the current slack plugin thats available. This plugin shows the notification count for the channels in their corresponding team as well as give you the option to mark either the channel as read or the whole team as read.


gingerbeardman:
Thanks!

PR #1208, opened 2019-03-29
Updated my plugins

phntxx:
Hello,

I'd like to update plugins I've previously pushed.
These plugins being:

- sbcDash-client: Updated it to work with the newer version of sbcdash
- pebble hearts: Pebble (along with the app store this plugin is meant to serve) doesn't exist anymore, so deleting this plugin seems like the right idea

Kind regards
Bastian Meissner


gingerbeardman:

PR #1220, opened 2019-04-16
initial release of my (e)GPU monitoring script

Rastafabisch:
It is writen with AMD Vega GPUs in mind, though it also works with Polaris. Intel is known not to work. I do not know about NVIDIA or older AMD GPU generations, though I suspect they work just fine. It is hardcoded to monitor the last GPU listet in Apple's System Profiler.app graphics section. This should be either your only GPU, your eGPU, if connected or on dual GPU MacBooks the dGPU. I don't know about, nor do not support hackintoshes, although they should work just fine.


Rastafabisch:
The build bot failed before setting up it's environment. Pulling should be fine. Working for me locally.

gingerbeardman:

PR #1233, opened 2019-05-16
Use threading in slack-unread

Kukant:
The previous implementation took my machine about 4.5 seconds to execute due to requests. By using threads to minimize the request waiting time, I managed to achieve under 0.8 second execution time.


notiv-nt:
Awesome 🖐

PR #1248, opened 2019-06-14
Update exchangeratesapi endpoint

eyupatis:
Previous endpoint was returning 404, this is the new endpoint of the api.


gingerbeardman:
Thanks!

PR #1255, opened 2019-06-19
Add missing files

benjifs:
I previously removed `package.json` as it was causing the tests to fail. Trying to add this back so its easier for people to run this

### DIFF
--- a/.test.py
+++ b/.test.py
@@ -7,6 +7,8 @@ import argparse
 import warnings
 from distutils.spawn import find_executable
 
+ignore_file_types = ['.md']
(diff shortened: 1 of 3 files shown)


gingerbeardman:
It's failing with 

GitHawk Upload by gingerbeardman

benjifs:
The latest commit adds a condition in the `test.py` script that skips `package.json` and `package-lock.json`. Not sure if that's allowed but it seems like the easiest way to deal with it. I've seen other bitbar plugins built with npm that have to create a shell script in order to replicate what the `package.json` file does.

gingerbeardman:
@benjifs let's try it!

PR #1271, opened 2019-07-22
Add Conditional For `None` Value in JSON

mtoconno:
There are certain conditions when the request returns data in the following format:

``` {u'quote': {u'symbol': u'[ticker]', u'changePercent': None, u'change': None, u'latestPrice': [price]}}}```

This adds a if statement to prevent getting an error symbol by printing the `latestPrice` value when the API interface returns `None` which is intepreted as a `null` value.


mtoconno:
After some more testing, I realized that negative values would return a `false` in the if statement.  Checking if the value `is not None` fixes this.

PR #1279, opened 2019-08-06
Progress Bar for Work Day (to beat procrastination!)

jasonrwang:
## Progress bar but for work day instead of 24h-day, incl. more features :)

Forked from [aurorabbit](https://github.com/aurorabbit) and [Mucahit](http://github.com/mucahit) (who submitted a [simpler version](https://getbitbar.com/plugins/Time/progress.1h.sh) w/ latest on 2156883).

Examples

### New Features

- [x] Lets user customize duration of day so it can show the work day instead (e.g. 09:00 to 23:00)
- [x] Gives different outputs to user based on the time – `P: xx%` during work time (`P:` to better indicate what it is), `🛑STOP!✋` when it's past the end of the day, and `😴SLEEP!🛌` when it's between midnight and the start of the new working day
- [x] Picked the San Francisco font for the menu bar output to blend in better

### New Features that need help!

- [x] Introduced colour glitch: works well with macOS Catalina Dark Mode, but drop-down menu progress colours are faded
- [x] Progress bar during `🛑STOP!✋` and `😴SLEEP!🛌` time show as >100% and negative, respectively
- [x] Width is probably not correct anymore... but it still works

### DIFF
--- a/Time/progress-procrastination.15s.sh
+++ b/Time/progress-procrastination.15s.sh
@@ -49,14 +49,14 @@ now=$(date +%s)
 
 Y=$(date +%Y)
 Y_start=$(date -j 01010000 +%s)
-Y_end=$(date -jr $Y_start -v +1y +%s)
+Y_end=$(date -jr "$Y_start" -v +1y +%s)
 Y_progress=$(
     echo "($now - $Y_start) * 100 / ($Y_end - $Y_start)" | bc -l
 )
 
 m=$(date +%m)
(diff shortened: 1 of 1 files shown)


gingerbeardman:
Thanks! Please fix Travis issues and I will happily merge.

jasonrwang:
🎉Fixed issues and passes Travis checks!

There were two issues previously. First one was me being lazy, second was syntactic!

gingerbeardman:
Bravo!

PR #1280, opened 2019-08-07
Fix url encoding bug

tomorrowkey:
# Bug

`DesktopCleaner#desktop_path` returns `Pathname`. but `Pathname` does not respond to `#gsub`.

# Solution

Use `#to_s` to be string before encoding.


gingerbeardman:
Thanks!

PR #1282, opened 2019-08-11
Pocket Bar: new large release

inbalboa:
* cache added
* keyring started to be used
* no more third party tools are needed to authorize the app
* more optimal API calls


gingerbeardman:
Thanks!

PR #1283, opened 2019-08-15
Fixed fan speed plugin

95gabor:
The fan speed plugin didn't work for me. I managed to fix it with same changes .

Please confirm that it works for you too.


gingerbeardman:
Thanks!

PR #1284, opened 2019-08-20
Add Switch Timezones plugin

Wasapl:
This plugin allows switch between several timezones. Zones and their names can be set in the $TZS array. Click on any timezone in the menu will switch timezone shown in the bar.
Click on Epoch, YMD, or ISO 8601 will copy current time to the clipboard.

### DIFF
--- a/Time/switchtimezones.5s.sh
+++ b/Time/switchtimezones.5s.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env zsh
+#!/bin/bash
 
 # <bitbar.title>Switch Timezones</bitbar.title>
 # <bitbar.version>v1.0</bitbar.version>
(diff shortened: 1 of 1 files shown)


gingerbeardman:
Please correct Travis CI errors so we can merge this PR

Wasapl:
Well, it's funny, but Travis error states that the shellcheck does not support ZSH scripts. So, the only way to fix this from my side is to rewrite it in BASH. I chose ZSH for better support of arrays and hash arrays. And one may find how tricky it is to get correct time zone from the array at line #39.
Also, I assumed that ZSH is fine for the community since [there](https://github.com/matryer/bitbar#tested-languages) is a statement "Anything that can write to standard out is supported".
But if you stand strong on "not to merge ZSH script" due to lack of support from the shellcheck then you should close this PR. Maybe there will be one day I feel inspired to rewrite this in BASH )))

gingerbeardman:
Thx! I appreciate your efforts on this.

PR #1294, opened 2019-08-30
add yubikey totp codes

brute-force:
-this uses yubikey manager cli https://github.com/Yubico/yubikey-manager to generate codes
-clicking on a menubar item copies that code to your clipboard


gingerbeardman:
thx!

PR #1298, opened 2019-09-07
fixed posix error no 2

mrlock13:
Fixed an error that'll happen on a machine that uses 2 or more version of python.
BitBar plugin will utilize the default system python version.


mrlock13:
tagging @eripa

gingerbeardman:
Thanks!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1299, opened 2019-09-09
Umbrella Status v1.2: Detect when re-enabled due to upgrade

jasmas:


gingerbeardman:
Please fix travis errors so we can merge

PR #1310, opened 2019-10-01
Add ISS Overhead Check plugin

mrprompt:
Added a simple script to show the moment when ISS pass ahead at of configured location.


mrprompt:
tks @gingerbeardman

PR #1312, opened 2019-10-01
add cloudwatch-alarms plugin

sebasrp:
Plugin that displays the CloudWatch alarms status in the region and account configured in boto3.
The dropdown will display the names of the alarms that are in the different statuses.

sample output:
cwa-plugin


sebasrp:
I messed up the PR as i forgot to split the 2 code changes into separate branches... will resubmit

Issue #1309, opened 2019-09-29
re: Battery remaining (Python)


iomatt512:
FAO: (@eripa)

Hey - your refresh interval shows incorrectly. Fixed it by moving the index for split back from 2 to 1:
Was:   
` refresh_interval = sys.argv[0].split('.')[2]`
Fixed:   
` refresh_interval = sys.argv[0].split('.')[1]`

Not sure how to pull/push this back to git so dropping it in here.

sebasrp:
this issue can be closed since the PR #1314  was merged

Issue #1130, opened 2018-11-16
re: Battery remaining (Python)


wayferer:
FAO: (@eripa)

I needed to add the following to the top of the file get things working right:

`#!/usr/bin/env PYTHONIOENCODING=UTF-8 /usr/local/bin/python3`

via https://github.com/matryer/bitbar/issues/141#issuecomment-171773836

Also, in order to get the interval to show correctly, I had to change:

`refresh_interval = sys.argv[0].split('.')[2]`

to:

`refresh_interval = sys.argv[0].split('.')[1]`

sebasrp:
this issue can be closed since the PR #1314  was merged

PR #1314, opened 2019-10-01
fix battery plugin and ensure we use python 3

sebasrp:
Fixes problems referenced in #1130 and #1309 by using the right split of the output and ensuring we use python3


sebasrp:
closing and re-opening to re-trigger a travis built

PR #1317, opened 2019-10-07
Add macOS 10.15 Catalina compatible Music plugin

jmunowitch:
The existing iTunes Lite plugin does not work with macOS 10.15. This version is compatible. 

Feel free to squash this PR into one commit.


gingerbeardman:
Please fix travis errors.

Does this change still retain working with legacy iTunes?

jmunowitch:
> Please fix travis errors.

Travis seems to have hung up in the setup phase, and I can't restart the build... Can you restart it?

> Does this change still retain working with legacy iTunes?

It does **not**. This is explicitly for 10.5 Catalina. The iTunes App was split into multiple apps, and the music version is now called `Music.app`.

gingerbeardman:
Have tried restarting it.

If this sheds compatibility with old iTunes it may be best having two separate plugins?

jmunowitch:
> Have tried restarting it.

Looks like the same error with Travis. It looks like there are a lot of build errors that are the same: https://travis-ci.org/matryer/bitbar-plugins/builds. It seems like the node version that it is upgrading to has changed, and the build is timing out on upgrade.

> Agreed that because this sheds compatibility with old iTunes it may be best having two separate plugins. Thanks!

I agree. This is a new plugin file called `music.10s.sh`. It is fully separate file from the existing iTunes plugin, even though the actual source code is almost identical, replacing `iTunes` -> `Music`

jmunowitch:
Oops! Thanks @tresni. Made the script executable and pushed the changes.

PR #1323, opened 2019-10-20
fix(googlecal): updates due to API changes and make dependencies exact

aibolik:


aibolik:
@kodie to approve this, as an author of the plugin

PR #1325, opened 2019-10-27
Extention of Daniel Seripap's weather plugin

JHale716:
An extension of Daniel Seripap's plug in after playing with the version from the BitBar website, rather than the slightly more polished GitHub version.

While Daniel Seripap's updated GitHub version has emoji support in it, I couldn't get the libary to load on my mac, so I've run with an embedded emoji approach.

Additionally with the Weather Condition fetched, localisation of the OpenWeather condition emoji mapping is included, with a couple of creative options. The core difference is the more severe weather has two emoji's rather than the vanilla single emoji.

### DIFF
--- a/Weather/weather-ext.15m.py
+++ b/Weather/weather-ext.15m.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# <bitbar.title>Weather - OpenWeatherMap</bitbar.title>
+# <bitbar.version>v1.0.2</bitbar.version>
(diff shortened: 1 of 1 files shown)


tresni:
Thanks for the PR @JHale716 . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs.  If you feel yours adds new functionality that should be shared, please do a pull request updating the existing weather plugin.  Thanks!

JHale716:
Thanks for the comments, fairly new here, so cautious of pushing things directly…
Happy to leave as is, if there’s something useful that can be used, cool, use it. If not, all good.

PR #1327, opened 2019-10-29
Add a musicRating script, compatible with MacOS 10.15 and older versions

WeibingChen17:


WeibingChen17:
The checks have failed but I don't think it is due to my new script. The build did not work

PR #1329, opened 2019-11-20
Add Discordian Date plugin

hpcodecraft:
Hi @matryer,

I've created a plugin to show the current Discordian Date. It would be awesome to see in in the official list, so here's a PR 😃


hpcodecraft:
❤️

PR #1331, opened 2019-11-23
Create jira-issues.10m.py

teopost:


tresni:
@teopost I wanted to close/reopen so we can check that it passes build test, but looks like the upstream repo has been deleted.  If you re-add/re-open this, we should be able to get it merged.

PR #1332, opened 2019-11-25
create W06 wifi transfer meter

token7:
see also : https://github.com/positrium/wifi-transfer-meter


token7:
@tresni Thank you . Then I fixed it and passed tests. I checked the operation and confirmed that there was no problem.

PR #1335, opened 2019-11-29
Drop outdated plugins for Dash cryptocurrency and add a new one

UdjinM6:
This drops outdated and no longer working plugins together with the folder they were in and adds a new one in the folder with the correct name.

Looks like this (in Light and Dark mode correspondingly):
Screenshot 2019-11-29 at 04 33 19
Screenshot 2019-11-29 at 04 33 43


tresni:

UdjinM6:
Rebased and refactored/fixed, should be ready now. I don't know if you prefer to squash&merge or to merge only. If it's the latter I can squash recent commits into b071d0e to keep `master` commit history clean.

Side question: if I would like to update other plugins I like e.g. add Dash to `CoinbasePrices.sh` or update Dash icon in `CryptoPrices.5m.sh` - what would be the best way to do it? Can I create a PR to do this myself (should I bump `<bitbar.version>` in this case?) or should I create an issue and wait for corresponding authors to update their plugins instead?

tresni:
Thanks for the update, I'll get this merged.  To answer your side question, either way is fine.  You are more than welcome to submit a PR with the changes you would like to see.  As long as you aren't putting it into a plugin it really doesn't belong with (e.g. a weather plugin), it's fine.  Please increment the version number when you do this and add yourself as the author.

Otherwise, you can open an issue and ask if the original author is interested in adding support.

PR #1336, opened 2019-12-01
Updated plugins to work with PagerDuty v2 API

pedrofuentes:
More details in https://github.com/pedrofuentes/bitbar-plugins


pedrofuentes:
@matryer it seems that the builds are failing because of an issue with CI upgrading python

PR #1341, opened 2019-12-13
Added nowplaying-modified.5s.sh

kb73:
A modified version of the nowplaying.5s.sh plugin by Adam Kenyon

### DIFF
--- a/Music/nowplaying-modified.5s.sh
+++ b/Music/nowplaying-modified.5s.sh
@@ -1,0 +1,149 @@
+#!/bin/bash
+
+# <bitbar.title>Now Playing Modified</bitbar.title>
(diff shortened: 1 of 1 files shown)


kb73:
There appears to be an issue with Travis CI in the .travis.yml file. This was the error:
 The command "gem install rubocop -v 0.50.0 --no-document" failed and exited with 1 during .

This appears to be an issue unrelated to my plugin. Can you confirm this?

tresni:
Thanks for the PR @kb73 . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs.  If you feel yours adds new functionality that should be shared, please do a pull request update the existing plugin.

PR #1343, opened 2019-12-16
Fixed the Irish Rail check

danburke:
Irish rail updated their API so this script broke. This version fixes this


tresni:

PR #1345, opened 2019-12-16
Add SureFlap cat flap/pet door plugin

henrik:
Intentionally uses Net::HTTP from Ruby's standard library, so you shouldn't need any further dependencies.


tresni:

henrik:
Thank you! :)

PR #1357, opened 2020-01-13
Fix Ruby 2.7.0 deprecation warning

mizoR:
Fixed following warn:

```rb
warning: calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open
```


tresni:

PR #1370, opened 2020-01-25
Add typerlc weather plugin

typerlc:
Add the plugin hosted here https://github.com/typerlc/bitbar-plugins to the Weather plugins


matryer:
Nice :)

PR #1378, opened 2020-02-05
added pianobarctl image, update mouse battery to use AppleDeviceManagementHIDEventService 

danlogan9:


System/Battery/mouse.1m.sh
```diff
@@ -8,7 +8,7 @@
 
 # works fine with Magic Mouse
 
-PERCENTAGE=$(ioreg -n BNBMouseDevice | fgrep BatteryPercent | fgrep -v \{ | sed 's/[^[:digit:]]//g')
+PERCENTAGE=$(ioreg -c AppleDeviceManagementHIDEventService -r -l | grep BatteryPercent | sed 's/[^[:digit:]]//g')
```
tresni: This looks like it would also pick up the battery percentage for a wireless keyboard.

danlogan9: I reverted the change on mouse.1m.sh.  I don't think the existing script works at all past Sierra through. I looked over what the AppleDeviceManagementHIDEventService returns and see that it could return other devices.  I am still using the old wired apple keyboard so I will resubmit something when I can work out just pulling the mouse percent.  thx

PR #1380, opened 2020-02-13
Add new coronavirus plugin

mrfhitz:


tresni:
Thanks for the PR @mrfhitz. We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs. There is an existing corona virus plugin at https://github.com/matryer/bitbar-plugins/blob/master/Environment/corona.1h.sh .  If you feel yours adds new functionality that should be shared, please do a pull request updating an existing price ticker.

PR #1394, opened 2020-03-12
SureFlap: Optimise and run more frequently

henrik:


henrik:
Added a commit to also handle network errors more gracefully.

tresni:
Toggling PR to trigger Travis build

henrik:
@tresni Good catch with deprecated “exists” vs “exist”. Thanks for merging!

PR #1395, opened 2020-03-13
fix: add music-controls support for Catalina Music.app

scottrbaxter:
Duplicated the iTunes references for all `music-controls-*.sh` plugins.

Tested and confirmed all are working in macOS Catalina (10.15.3)


tresni:
Generally we like to see version increments, but I'll merge this as it adds functionality, doesn't change it.

PR #1396, opened 2020-03-14
Cryptocurrency: Dash: Replace Coinmarketcap.com api with the one from Coincap.io, add Binance as an alternative to Poloniex

UdjinM6:
Coinmarketcap stopped providing (unlimited) free API access unfortunately, have to use another data source. Coincap.io looks like a good choice for now. Also adding Binance.com as a more trading volume/less bandwidth consumption alternative to Poloniex.com (API of the latter is also failing sometimes for some reason).

coincap_poloniex
coincap_binance

### DIFF
--- a/Cryptocurrency/Dash/dash-coincap-poloniex.5m.sh
+++ b/Cryptocurrency/Dash/dash-coincap-poloniex.5m.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-# <bitbar.title>Dash tickers: Coinmarketcap and Poloniex</bitbar.title>
+# <bitbar.title>Dash tickers: Coincap.io and Poloniex.com</bitbar.title>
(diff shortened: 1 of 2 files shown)


Cryptocurrency/Dash/dash-coincap-poloniex.5m.sh
```diff
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-# <bitbar.title>Dash tickers: Coinmarketcap and Poloniex</bitbar.title>
+# <bitbar.title>Dash tickers: Coincap.io and Poloniex.com</bitbar.title>
 # <bitbar.version>v1.0</bitbar.version>
```
tresni: Please increment the version so people know this is a change.

UdjinM6: Done

PR #1397, opened 2020-03-14
Add AWS EC2 instance start/stop plugin

franfran:
Hi,

This is a new plug-in for you to start/stop EC2 instance.

It will load your instance list from your AWS profile automatically.

Thx.

### DIFF
--- a/AWS/ec2-start-stop.1h.sh
+++ b/AWS/ec2-start-stop.1h.sh
@@ -1,0 +1,83 @@
+#!/bin/sh
+
+# <bitbar.title>Amazon EC2 Instance Start/Stop</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Francis Mak</bitbar.author>
+# <bitbar.author.github>franfran</bitbar.author.github>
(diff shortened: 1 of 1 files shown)


franfran:
Hi,

My CI failed due to single quote issue in JQ command.  However, I can see other users can do pull request with single quote in JQ without problem.

Anyway to fixed this?

Thx.

franfran:
Never mind, it has been fixed now, thx.

tresni:
If you want to use `$CMD_JQ`, you should be able to do `#shellcheck disable= SC2016` on the line preceeding.  My guess is shellcheck has some logic to detect certain commands and understands they have variables using `$`.

If you are okay with it as it is, I'll merge.

franfran:
Thx @tresni for checking my code.
I am ok with this, pls merge.

FYI, it seems shellcheck look against the followings commands(including the jq)
https://github.com/koalaman/shellcheck/blob/a57f6d2886e899133218e238fb1a3b26f948010d/src/ShellCheck/Analytics.hs#L978

gabrielcosi:
Hi, I'm new to aws cli, I've installed and configured it with my private and access keys, jq is already installed but I'm getting an error jq: error (at <stdin>:120): Cannot iterate over null. What am I doing wrong?

PR #1405, opened 2020-03-29
Add covid-19_stats.15m.sh

wilsongoode:
Adding a new plugin I created to track COVID-19 in the United States.

### DIFF
--- a/Environment/covid-19_stats.15m.sh
+++ b/Environment/covid-19_stats.15m.sh
@@ -1,86 +1,145 @@
-#!/bin/zsh 
+#!/bin/zsh
 
 # <bitbar.title>COVID-19 Stats</bitbar.title>
-# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.version>v1.1</bitbar.version>
 # <bitbar.author>Wilson Goode</bitbar.author>
 # <bitbar.author.github>wilsongoode</bitbar.author.github>
 # <bitbar.desc>Displays stats of US COVID-19 cases, with a submenu for user-defineable States. Can also be configured to show the top n states.</bitbar.desc>
 # <bitbar.image>https://github.com/wilsongoode/covid-bitbar/blob/master/screens/covid-19_top15_states.png</bitbar.image>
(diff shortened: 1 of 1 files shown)


tresni:
Toggling PR to trigger Travis build

Environment/covid-19_stats.15m.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/zsh
```
tresni, line 1: Please use /bin/bash or another supported shell.  zsh scripts are not supported as we can't lint them.

Environment/covid-19_stats.15m.sh
```diff
@@ -0,0 +1,145 @@
...
+
+# <bitbar.title>COVID-19 Stats</bitbar.title>
+# <bitbar.version>v1.1</bitbar.version>
+# <bitbar.author>Wilson Goode</bitbar.author>
+# <bitbar.author.github>wilsongoode</bitbar.author.github>
+# <bitbar.desc>Displays stats of US COVID-19 cases, with a submenu for user-defineable States. Can also be configured to show the top n states.</bitbar.desc>
+# <bitbar.image>https://github.com/wilsongoode/covid-bitbar/blob/master/screens/covid-19_top15_states.png</bitbar.image>
```
tresni, line 8: Use the raw link of https://github.com/wilsongoode/covid-bitbar/raw/master/screens/covid-19_top15_states.png

Environment/covid-19_stats.15m.sh
wilsongoode: updated the shebang and needed to update one addition line because of a change to the behavior of grep. Thanks for the tip, I'll make sure any future scripts are bash from the start. As zsh is now the default shell, would zsh linting be possible in the future? I'm not familiar enough to know why it wouldn't be lint-able

Environment/covid-19_stats.15m.sh
wilsongoode: thank you, updated!

tresni:
Cleaned up a couple lingering shellcheck issues in 8d99068

PR #1408, opened 2020-03-31
add ipinfo plugin to Networking

j4ng5y:
Please see https://github.com/j4ng5y/ipinfo-bbp.git for source code.
I wasn't entirely sure how to give you that without just linking it lol.


tresni:
Hi @j4ng5y , we don't support binary plugins in this repository at this time.  Please see [Network/checkhosts.1m.go](https://github.com/matryer/bitbar-plugins/blob/ec3ac184198936271503018330314fa38ad63a1a/) for an example of how to add a go plugin using gorun.

PR #1409, opened 2020-03-31
Fix travis

j4ng5y:
There is no @2 version of python in the homebrew mainline anymore. If this is still needed, we would need to install it another way.


tresni:
Sorry for the delay.  Fixed in 8d581ac6bdd139bfd5f8e9022caee0bdb08b0b61

PR #1412, opened 2020-04-03
Add openfortivpn Bitbar plugin

emranovin:
- Add a plugin to be able to connect/disconnect or track openfortivpn status by using fortivpn-config certificate file.


emranovin:
I've fixed the issues.

PR #1413, opened 2020-04-04
Create pastebin.1m.sh

LivelyCarpet87:
Created a bash script that queries the PasteBin API for a list of pastes of the user. It would provide links to the pastes it detects.


tresni:
attempting to re-trigger travis build

LivelyCarpet87:
Hello, and thanks for the feedback. I have already corrected the errors raised in the last pyflake run except SC2206, which is intentional to split the input into an array. I have also added more error handling and an image.

tresni:
@LivelyCarpet87  Please review the changes in d068111 which should address the last of your linter issues.  If you are good with my changes, simply comment here and I'll merge everything.

LivelyCarpet87:
Thank you so much for digging through all my spaghetti bash script and fixing all my sloppy bugs. Everything seems good to go. Again, thank you so much.

PR #1415, opened 2020-04-07
Adding YNAB budget tracker

noahsphillips:


tresni:
A fellow YNABer ;) My only problem is that we haven't generally had "installer" plugins in this repository.  It would be appropriate to include the .js version of your script here.  It's understood that there may be dependancies to install to get a script to work.

sprak3000:
@noahsphillips 

Checking in to see if you are still maintaining this plugin. If you are, these changes would need to be made:

1. Rather than a bash script to install the plugin, the PR should be adding the actual [JS plugin](https://github.com/noahsphillips/ynab-bitbar/blob/master/ynab-bitbar.15m.js).
2. The JS plugin should have the metadata tags in it, updating the names from `<bitbar...>` to `<xbar...>`.
3. The image metadata tag needs to link to a screenshot of the plugin in action.

PR #1416, opened 2020-04-08
Remove docker-machine dependency

sudocurse:
Since Docker for Mac is [no longer packaged with `docker-machine`](https://github.com/docker/for-mac/issues/4208#issuecomment-576811681), docker-machine should not be a dependency; currently an up-to-date install produces an error for `docker-machine: command not found`. 

This removes the docker-machine requirement by preventing the variables from being set if the commands don't exist. This way the [empty string conditional](https://github.com/matryer/bitbar-plugins/blob/7bddc2f1692a80c774535f6aad18d2c2f45c2403/Dev/Docker/docker-status.1m.sh#L49) works as intended, and the `[ -n ... ]` tests don't return true when they shouldn't.


sudocurse:
Sidenote, [`dlite` was abandoned months after the last commit](https://github.com/veggiemonk/awesome-docker/commit/e6edc5d9a2aa4c3f32b42320cdbff36aa9fdfcdf) and I'm not sure modern Docker for Desktop needs it.

tresni:
Cleaned up a few shellcheck issues with b14acf7 .  I use docker native so tested with that.  Also, incremented version number.

PR #1419, opened 2020-04-11
Toggle Handoff, tested on macOS Catalina

martinschilliger:


tresni:
We use shellcheck to lint shell plugins.  There were a couple minor issues that I cleaned up in 
6c831d0.  Please review that for future reference!

martinschilliger:
Nice, thank you!

PR #1433, opened 2020-04-30
fix: update to python3

tgrrr:
urllib2 -> urllib
docs: 
https://stackoverflow.com/a/2792652/3281978
https://stackoverflow.com/a/28906913/3281978


Dev/RescueTime/rescuetime.1h.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env PYTHONIOENCODING=UTF-8 # /usr/local/bin/python3
```
tresni, line 1: This shebang means the script won't run correctly..  I believe you might be looking foir:

```
#! /usr/bin/env PYTHONIOENCODING=UTF-8 python3
```

Issue #1432, opened 2020-04-29
re: Amazon EC2 Instance Start/Stop


gabrielcosi:
FAO: (@franfran)
<img width="406" alt="Screen Shot 2020-04-29 at 13 04 05" src="https://user-images.githubusercontent.com/44176694/80630488-e9d3aa80-8a19-11ea-9482-ad7e1de79c09.png">

franfran:
Hi, can you run the aws command and see if you have the expected json output:

`aws --profile <your_profile> ec2 describe-instances --output json > ~/sample.json`

In the output file ~/sample.json, you should see structure like below.  If not, maybe you have no instance?  Wrong region in your profile?

```json
{
    "Reservations": [
        {
            "Instances": [
                {
                    "AmiLaunchIndex": 0,
                    "ImageId": "ami-123",
                    "InstanceId": "i-123",
                    "InstanceType": "t3.nano",
                    "KeyName": "<123>",
                    "LaunchTime": "2020-04-27T14:19:06+00:00",
                    "Monitoring": {
                        "State": "disabled"
                    },
                    "State": {
                        "Code": 80,
                        "Name": "stopped"
                    },
                    "Tags": [
                        {
                            "Key": "Name",
                            "Value": "optional-server-name"
                        }
                    ]
                }
            ]
        }
    ]
}
```

franfran:
Hi,

After running the aws command, can you confirm if there is a "Tags" in your json?
Usually we use Tags Name to identify multiple servers, but maybe this is not suite for your case.

If there is no "Tags" at all, please try:

1. edit the ec2-start-stop script
2. on line 60, change the jq command 
from
`if count (.Tags[] | select(`
to
`if count (.Tags[]? | select(`
3. make sure you save and restart the bitbar

Pls help to confirm if the above works, then I can fix this bug and do another pull request.

Thx!

gabrielcosi:
That fixed the problem, thanks!

gabrielcosi:
That fixed the problem, thanks!

PR #1435, opened 2020-05-02
Bug fixed for #1432

franfran:
-bug fixed for ec2 without Name tag
-added shellcheck to skip jq command


tresni:
You need to rebase your change on master to avoid the conflicts or you may be able to deal with it through the Github UI.

franfran:
I fixed the conflict using Github UI instead, thx.

PR #1440, opened 2020-05-11
Create weather.1h.py

JAckZ97:


tresni:
Thanks for the PR @JAckZ97 . We really try to have 1 example plugin that does one thing and allow other to modify it to fit their needs. There are a number of weather tracker plugins already, including one for OpenWeatherMap (Weather/OpenWeatherMap/weather.15m.py.)  Can you explain how yours is different from the existing plugin?  Is it functionality that could be added to the existing plugin?

Thanks!

JAckZ97:
Thanks for reviewing. Yeah at the beginning I try to use the one for OpenWeatherMap (Weather/OpenWeatherMap/weather.15m.py.) plugin, but I find it has dependence of emoji module. Which means as A new user, when I just download I can not use it right away. So I change the code and add more features like feels like, min/max temperature, humidity, wind etc. And the UI looks clearly.

sprak3000 (approved):

PR #1444, opened 2020-05-21
Add ejector plugin for ejecting/unmounting volumes

Gruntfuggly:


Gruntfuggly:
Lol - I've just found another Ejector plugin in the System folder. Mine is much simpler and only shows up when there are volumes mounted. I'll rename it and you can decide whether it should be included or not.

PR #1448, opened 2020-05-25
add fan_speed plugin for newer model

tamanishi:
There is already a plugin that shows fan speed, but it doesn't work on recent models.
This pull request adds a plugin for recent models.

### DIFF
--- a/System/fan_speed.5s.js
+++ b/System/fan_speed.5s.js
@@ -4,7 +4,7 @@
 // <bitbar.author>Masayuki Sunahara</bitbar.author>
 // <bitbar.author.github>tamanishi</bitbar.author.github>
 // <bitbar.desc>Shows fan speed. Strongly inpired by Eric Ripa's "Fan Speed" plugin.</bitbar.desc>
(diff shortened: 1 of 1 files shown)


System/fan_speed.5s.js
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env /usr/local/bin/node
+// <bitbar.title>fan_speed</bitbar.title>
+// <bitbar.version>v1.0</bitbar.version>
+// <bitbar.author>Masayuki Sunahara</bitbar.author>
+// <bitbar.author.github>tamanishi</bitbar.author.github>
+// <bitbar.desc>Shows fan speed. Strongly inpired by Eric Ripa's "Fan Speed" plugin.</bitbar.desc>
+// <bitbar.image>https://github.com/tamanishi/fan_speed/blob/master/image.png</bitbar.image>
```
matryer, line 7: please use `https://github.com/tamanishi/fan_speed/blob/master/image.png?raw=true`

tamanishi:
Thank you for reviewing.
I fixed the image url.
Please check.

sprak3000 (approved):

PR #1464, opened 2020-07-03
Add plugin for Pi-hole

alst-ovh:
Add plugin for Pi-hole

dw-pihole

Plugin Repo: https://github.com/alst-ovh/bitbar-dw-pihole

### DIFF
--- a/Network/dw-pihole.10s.py
+++ b/Network/dw-pihole.10s.py
@@ -5,7 +5,7 @@
 # <bitbar.author>Albert Steiner</bitbar.author>
 # <bitbar.author.github>alst-ovh</bitbar.author.github>
 # <bitbar.desc>Show Summary and Manage Pi-hole from Menubar.</bitbar.desc>
-# <bitbar.abouturl>https://github.com/alst-ovh/bitbar-dw-pihole/blob/master/dw-pihole.png</bitbar.abouturl>
+# <bitbar.abouturl>https://user-images.githubusercontent.com/35896287/86492400-5480c980-bd6e-11ea-83c1-081a11c3d63c.png</bitbar.abouturl>
 import json
 import os
 import ssl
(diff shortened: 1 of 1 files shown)


Network/dw-pihole.10s.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# <bitbar.title>Pi-hole Status</bitbar.title>
+# <bitbar.version>v1.0</bitbar.version>
+# <bitbar.author>Albert Steiner</bitbar.author>
+# <bitbar.author.github>alst-ovh</bitbar.author.github>
+# <bitbar.desc>Show Summary and Manage Pi-hole from Menubar.</bitbar.desc>
+# <bitbar.abouturl>https://github.com/alst-ovh/bitbar-dw-pihole/blob/master/dw-pihole.png</bitbar.abouturl>
```
matryer, line 8: Could you add the image? I think `https://user-images.githubusercontent.com/35896287/86492400-5480c980-bd6e-11ea-83c1-081a11c3d63c.png` will do.

jebabin:
There is already a pi-hole plugin in the directory, named pi-hole, why not doing a pull request on it has your code seems an updated version of the existing code ?
Also see my pull request #1600 you could reuse the part where I put base_url as a xbar variable.
I'm also wondering why you refresh every 10s by default ? This seems very high, may be some use case may need this, but for me, a default to 10m would be better for most user to save ressources

matryer:
@alst-ovh have you seen the existing plugin? What would you like to do?

alst-ovh:
I now the pi-hole plugin but not working for me so i forked the new one, added some extensions to it and for these it should refresh every 10s.
Work's for me, i wont do anything to it now, if you don't wont it, let it be ...

matryer:
@alst-ovh I'd like to replace the broken one with yours. Would you be able to do a PR that preserves the previous author's name? Also, would you be able to add `xbar.image` to the metadata linking to the image above?

PR #1510, opened 2020-10-03
Fix 'no screens' state in screen.2m.sh plugin

Dispader:
The conditional that checks for a 'no screens' state was erroneously only checking for a lack of return from 'screen -list'.  A successful 'screen -list' contains an 'is empty' message which prevented the
display.

This commit also adds an action to create a screen session if none currently exist.  So here we change the 'no screen' message to 'new screen session'.

Screen Shot 2020-10-03 at 3 24 43 PM

### DIFF
--- a/Tools/screen.2m.sh
+++ b/Tools/screen.2m.sh
@@ -14,8 +14,8 @@ PATH="/usr/local/bin:$PATH" SCREEN_COMMAND=$(command -v screen)
 echo "💻"
 echo '---'
 SCREENS=$(${SCREEN_COMMAND} -list | grep -o '\s*.*\s*(.*)')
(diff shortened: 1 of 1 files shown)


Dispader:
It looks like the `bitbar-plugins` Repository isn't participating in [Hacktoberfest](https://hacktoberfest.digitalocean.com/).  😢 

Screen Shot 2020-10-03 at 3 27 22 PM

Any change you'd consider adding the `hacktoberfest` topic to `bitbar-plugins`?

It's not a huge deal— I'm happy with this small fix (already installed on my system) and think the Pull is a nice improvement that I'd like to get out to plugin users; but it'd be nice to get a little gamification love in October.

PR #1536, opened 2021-01-02
Add sunrise plugin

bgreenlee:


matryer:
@bgreenlee makes me want coffee. Nice work.

PR #1539, opened 2021-01-08
add simple python based octoprint job percent

tavdog:
This is a very simple current job percent displayer for octoprint in python.  The existing octoprint plugin is harder to user and configure as a shell script. This python version is a minimal job percent display.

### DIFF
--- a/Tools/octoprint_job_percent.1m.py
+++ b/Tools/octoprint_job_percent.1m.py
@@ -11,10 +11,10 @@ import requests
 import re
 
 # edit to your octoprint ip address:port
-url = "http://192.168.1.39:5000/api/job"
+url = "http://octopi.local/api/job"
 
 # update to your api-key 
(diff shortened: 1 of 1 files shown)


matryer:
@tavdog Thanks for this - I wonder if you could take a screenshot of it running and include the image metadata?

tavdog:
Well it's just a percentage number or a dash when not printing.  What image metadata do you mean ?

zakx:
@tavdog I guess the meaning was for you to create a screenshot and include it as `<bitbar.image>` in your plugin's metadata. :)

matryer:
@tavdog sorry to bother you, but that image is really gigantic. I wonder if you could replace it with a smaller more focussed screenshot of the menu itself, much like you see on xbarapp.com.

tavdog:
Oh wow, that was really big.  No problem.  I've cropped the image way down.

PR #1540, opened 2021-01-12
new Plugin: CO2 emissions for producing electric energy

pygoner:
- plugin displays the current carbon (gC02equivalent) emissions per kWh of produced electric energy in the requested country/region
- uses CO2 Signal API


matryer:
@pygoner nice one - I wonder if you could take a screenshot of the thing and add the image metadata?

matryer:
@pygoner sorry for not being clear. The binary doesn't go in this repo, at the moment they're elsewhere - the metadata `<bitbar.image>` should be a URL that points to the image. (See other plugin examples)

pygoner:
@matryer sry, this was a mistake by me. Now the image link should be in the right manner. Thank you for your response.

PR #1546, opened 2021-02-03
Add pidkiller plug-in to plugin repository

dengel29:
Made a plug-in which can display all running processes + process IDs (PIDs) and allows you to kill any upon clicking them.


dengel29:
Something went wrong, will resubmit with correct file

PR #1548, opened 2021-02-05
Fix GitHub contributions regex after markup change

rexxars:
GitHub seems to have changed the markup for their contributions "partial", so the regex does not match anymore.
I've changed the regex slightly, and also had to add a break to the scanning loop as it was continuing to the end of the year - I think maybe they've removed a feature of the endpoint so that it doesn't stop on the date you give it.


matryer:
@rexxars I think this has already been fixed now?

rexxars:
Looks like there's been a few other PRs merged for this, yeah!

PR #1552, opened 2021-02-11
Weather source update

seripap:
Not sure why this was replaced, would like to keep original credits for my work.


matryer:
@seripap strange - sorry about that. Perhaps we need to support multiple contributors.

PR #1557, opened 2021-03-02
BMW Charging Status plugin

zakx:
Hi,

this plugin will log into BMW's ConnectedDrive platform and display your car's charging status and ETA (while charging) or your remaining range.

Screenshot 2021-03-02 at 02 09 32


matryer:
@zakx what a great plugin, thanks

PR #1558, opened 2021-03-02
slight restructure

matryer:
I've removed any top-level categories that only contain a single plugin.


matryer:
@melonamin Do you need some notice before structural changes like this happen?

melonamin:
Hi @matryer if you changing repo structure than yes, things will break. A notice wouldn’t help me, because I need the change to happen before I can adopt it.

So I would say - don’t bother, I’m following changes in this repo, and I’ll pick up what ever is important to me. Thanks for asking!

PR #1572, opened 2021-03-21
Add a Yubikey plugin

jebabin:
Plugin that generates totp codes using yubikey and copy to clipboard, supports touch and touchless code generation.
Use ykman from /Applications/YubiKey Manager.app


matryer:
@jebabin thanks for this plugin - could you change `bitbar` metadata to `xbar`? I've updated the test script so it should pass if you do so.

matryer:
retest this please

jebabin:
Done

PR #1575, opened 2021-03-28
Update netstats.5s.sh with xbar variables

Joshfindit:
Also tweaks the logic so it's more robust and hides wifi stats when there is no wifi.


Joshfindit:
(Fixes https://github.com/matryer/xbar-plugins/issues/1512)

matryer (approved):

PR #1585, opened 2021-04-03
Creation of a Broken directory for broken plugin and first move in

jebabin:
This is a proposal to move broken plugin in a specific directory, so that users wont try to install and not understand why it's not working and dev who want to fix plugin know where to focus on. 
This include the move of a first plugin coin-alert who rely on a no longer existing API


matryer (approved):

matryer:
@jebabin I'm not sure we want a Broken folder, because the category becomes "Broken" on the website. But I agree this is a problem that we want to solve.

matryer:
@jebabin how about we do `<xbar.working>true</xbar.working>` - and only add that to working plugins?

jebabin:
Yes make sens

jebabin:
Thinking about it, if you only add <xbar.working>true</xbar.working> to working plugins, do you intend to update all plugins with this by default ? The idea of a Broken directory was also that you can hide it in the GUI.
So I think that <xbar.working>false</xbar.working> should be added to broken plugins so that you can filter them out of the GUI

jebabin:
I opened an issue https://github.com/matryer/xbar-plugins/issues/1593 to discuss this

PR #1598, opened 2021-04-13
Update Foobot.10b plugin to use xbar.var for API_KEY and DEVICE_ID

alessio-signorini:


Weather/Foobot.10m.rb
```diff
@@ -17,11 +17,11 @@
 
 # Request API key from
 #    http://api.foobot.io/apidoc/index.html
-API_KEY=" ... "
+# <xbar.var>string(API_KEY=""): API Key</xbar.var>
```
matryer, line 20: tip: If you use `VAR_API_KEY` and `VAR_DEVICE_ID` you'll get better labels in the UI.

If those env vars are standard or something, then maybe it's better to leave them?

matryer: @alessio-signorini let me know what you want to do, I'm happy to merge it as it is.

alessio-signorini: Thanks for the tip, @matryer. I would merge as is. It's a click for both of us and it doesn't make a big difference.

I'll improve it at the next update!

sprak3000 (approved):

PR #1602, opened 2021-04-17
Fixes unquoted font name string causing 'malformed parameters: missing equals' error

cbergoon:
The PR adds quoting to a font name in Time/progress-procrastination.15s.sh that has spaces causing xbar to be unable to parse the output. Attached is the original error caused by this plugin: 

 
progress-xbitbar-issue


matryer:
@cbergoon thanks for this. We should make sure there's an issue open in https://github.com/matryer/xbar to fix this too.

PR #1624, opened 2021-05-21
Add KSing plugin

xfangfang:
KSing is a karaoke mobile app from China, which allows users to sing and upload their own songs to share with others, but unfortunately there is no way to listen to them on the Mac.
This plugin helps people who want to listen to music from KSing on the Mac. You can click the help in the plugin to see the detailed usage. 
Later, I will give a readme of the English Version (although I don't think there may be any English users)


matryer:
@xfangfang nice job

PR #1644, opened 2021-06-20
Update digitalocean.1h.js

carstn:
When formatting a droplet's `created_at` Date object into a string, the month is not formatted properly (due to missing parentheses). Currently, the string "1" is concatenated to the zero-indexed month string. Instead, we need to add the integer 1 to the zero-indexed month returned by `date.getMonth()`. Adding parentheses around `date.getMonth() + 1` returns the sum of the two integers instead of concatenating the two strings.
fao @hasit


carstn:
See issue https://github.com/matryer/xbar-plugins/issues/1643

PR #1646, opened 2021-06-29
Fix diskAvailable plugin raising error window on BigSur

luka-papez:
The plugin runs "diskutil info" command and for me the output of the command says "Container Free Space" and not "Volume Free Space" (as the plugin expects).
I am running BigSur 11.2.3 and the plugin was raising an error window saying "Application is not open anymore".

This small tweak fixes the problem and it looks to me that with this tweak the plugin should work on both diskutil outputs ("Container" and "Volume").
Sadly, I have no way to test this on an actual machine that outputs "Volume" and some feedback would be much appreciated.


matryer (approved):
Seems sensible.

PR #1649, opened 2021-07-01
Another plugin for ethereum gas price

ichuan:
Show current ethereum gas price (fast and standard) in Gwei.
Data source: www.gasnow.org


matryer:
@ichuan this is cool, thanks for writing it - is there another plugin that already does this? We should probably try to stick to just one, if so. But if you think they're different enough, could you update the image metadata to point to that screenshot? It'll fit in better with the others.

ichuan:
@matryer Hi, I find out there's already a `gasnow.10s.sh` does the same thing. Closing this issue now.

PR #1658, opened 2021-07-10
Add feature-filled Nagios plugin

seren:
This plugin outputs Nagios status.

It's based on the `nagios.30s.sh` plugin by Rob DeSanno (@rdesanno), but improves on it in the following ways:
- Uses xbar variables for credentials and settings (no script editing required)
- Is compatible with Nagios servers that live behind a VPN.
- Output has been significantly (IMO) improved.

Screen Shot 2021-07-10 at 12 24 44 AM

Screen Shot 2021-07-10 at 12 34 41 AM

### DIFF
--- a/Dev/Nagios/nagios.30s.sh
+++ b/Dev/Nagios/nagios.30s.sh
@@ -1,38 +1,126 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# <xbar.title>Nagios Prod</xbar.title>
+set -uo pipefail
+
+# <xbar.title>Nagios</xbar.title>
 # <xbar.version>v1.0</xbar.version>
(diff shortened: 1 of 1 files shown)


matryer (changes requested):
@seren if this replaces another plugin, we should overwrite the file (or remove it) and include @rdesanno in the metadata as an author. What do you think?

seren:
That sounds fine. I didn't want to mess with someone else's plugin, but it makes sense not to duplicate work. Is there a way to add multiple authors, or should I just specify @rdesanno or myself?

I've updated the PR to overwrite the existing script rather than adding a new one.

sprak3000 (approved):

PR #1659, opened 2021-07-10
Add Pagerduty plugin

seren:
This plugin will display open Pagerduty incidents assigned to a user.

Screen Shot 2021-07-10 at 12 19 11 AM

### DIFF
--- a/Tools/PagerDuty/pagerduty_incidents.1m.sh
+++ b/Tools/PagerDuty/pagerduty_incidents.1m.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# <xbar.title>Pagerduty Incidents</xbar.title>
+# <xbar.version>v1.0</xbar.version>
+# <xbar.author>Seren Thompson</xbar.author>
+# <xbar.author.github>seren</xbar.author.github>
+# <xbar.desc>Displays the number of triggered Pagerduty.com incidents assigned to a user</xbar.desc>
+# <xbar.image>https://imgur.com/LJR06et.png</xbar.image>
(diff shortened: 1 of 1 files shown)


Dev/pagerduty.1m.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# <xbar.title>Pagerduty Incidents</xbar.title>
+# <xbar.version>v1.0</xbar.version>
+# <xbar.author>Seren Thompson</xbar.author>
+# <xbar.author.github>seren</xbar.author.github>
+# <xbar.desc>Displays the number of triggered Pagerduty.com incidents assigned to a user</xbar.desc>
+# <xbar.image>https://i.imgur.com/2p2nFg8.png</xbar.image>
```
matryer, line 10: Please use an image that shows the plugin when it's installed.

matryer (changes requested):
There are already some PagerDuty plugins, please ensure this doesn't duplicate behaviour.

seren:
I don't think this duplicates behavior of the other three existing plugins, though `pagerDutyIncidents.30s.js` might be the closest. I'm unable to run any of them out of the box (fresh install of Mac OS 11.5.2) however:
- `pagerduty.30s.rb` fails to run due to missing ruby gem dependencies.
- `pagerDutyOnCall.30s.js` and `pagerDutyIncidents.30s.js` require node and a `.bitbarrc` config file, neither of which I have installed/created.

This version uses xbar variables and is a simple shell script, so it should work out of the box and be easy to configure via the xbar gui.

I've updated the image to show the plugin's menu.

sprak3000 (approved):

PR #1661, opened 2021-07-12
Remove life-percentage.12h.rb

krmbzds:
I created this plugin long time ago.
There is a newer plugin with equivalent functionality so I'd like to remove this one.


matryer:
@krmbzds this is the first PR of this kind and I can't thank you enough. I love PRs that delete code. Please let me know if you'd like to delete anything else :)

PR #1667, opened 2021-07-27
add location-switcher plugin

aaronellington:


Network/location-switcher.5s.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# <xbar.title>location-switcher</xbar.title>
```
matryer, line 3: Please add an image of the plugin when it's open.

matryer (changes requested):

sprak3000:
@aaronellington 

Poking about getting an image in place in the metadata.

PR #1669, opened 2021-07-29
Add plugin to toggle wifi when laptop lid gets opened/closed

manuel-koch:


Network/wifi-toggle-by-lid.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+#
+# Switch off Wifi when laptop lid is closed and force power-management-state sleeping.
+# Switch on Wifi when laptop lid is opened again.
+#
+#  <xbar.title>wifi-toggle-by-lid</xbar.title>
```
matryer, line 6: Please add an image showing the plugin when it's installed.

matryer (changes requested):

Network/wifi-toggle-by-lid.sh
manuel-koch: I added a "<xbar.image>" entry to the comment section of the script.

manuel-koch: This is my first Pull Request - I'm not sure whether "starting a review" in Github was the correct reaction to your comment ?
I updated my branch and added an image for the plugin.

sprak3000 (approved):

PR #1677, opened 2021-08-09
Add IPs plugin to show local and external IP

martin-braun:
Plugin to allow displaying and copying the local and external IP address of the device.

### DIFF
--- a/Dev/ips.30s.sh
+++ b/Dev/ips.30s.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Display local and external IP and allow copying it. This plugin will connect to icanhazip.com to determine external IP address.
+#
+# by Martin Braun (martin-braun.net)
+#
+
+# metadata
+# <xbar.title>IPs</xbar.title>
+# <xbar.version>v1.0.0</xbar.version>
+# <xbar.author>Martin Braun</xbar.author>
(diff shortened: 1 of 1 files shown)


matryer:
Hey @martin-braun - how different is this to https://xbarapp.com/docs/plugins/Network/netstats.5s.sh.html?

martin-braun:
@matryer Thanks for checking out my script. I didn't recognize netstats that looks promising. Although, here are some points that I think it lacks: 

- Showing the local IP right in the menu bar
- Allowing to click the IPs to copy them to clipboard in the menu
- On my machine it can only resolve the WAN-IP, not the LAN-IP

Although there implementation works in my Terminal: 

```sh
$% route -n get default | grep -o "interface: .*" | sed -e 's/interface: //'
en12
$% ipconfig getifaddr "en12" 2>&1                    
192.168.138.243
```

It does not in xbar: 

Screen Shot 2021-09-01 at 20 44 29

I don't know why, right now.

I probably could enhance this existing script to allow copying of IPs, but I really have benefit in seeing the latter part of my local IP in the menu bar. I would not want to change this behavior, because it would change the experience for existing users and is purely subjective.

I like the idea of the 3 clickable options in the menu, but nothing happens when I click them.

What would be your recommendation?


macOS BigSur 11.5.2 (20G95)

sprak3000 (approved):

Issue #1688, opened 2021-09-05
Remove unnecessary absolute path references


kidonng:
xbar is great, however plugin quality varies. In particular, there are many scripts that reference absolute paths unnecessarily or don't checking if it exists.

Unnecessary absolute path references include programs already presented in `$PATH`:

- **`/usr/local`, in particular `/usr/local/bin`**
- `/usr/bin`
	- `/usr/bin/env` is an exception
	- `/usr/bin/python`, `/usr/bin/ruby`: these are scheduled to be removed, but it is probably too much a hassle to check for other available Python or Ruby runtime so it's ok to keep them until the day they _really_ break
	- `/usr/bin/php` is already removed in macOS 12 (Monterey)
- `/bin`: `/bin/sh`, `/bin/bash` and `/bin/zsh` should be ok. I checked with `rg '#!/bin' | rg -v '/bin/(bash|zsh|sh)'` and get no results so this one should be fine
- `/usr/sbin`
- `/sbin`

I would particularly note that `/usr/local` is essentially the default Homebrew prefix for Intel Macs, it is not only **incompatible with system not using Homebrew**, but also **incompatible with the latest and all future ARM Macs** (as default Homebrew prefix is `/opt/homebrew` there).

There is a slight slowdown of switching those references to `/usr/bin/env <program>` but it should be neglectable.

Another exception is that changing `$PATH` to include those paths is also fine.

---

I'd like to propose two changes regarding this issue:

- Remove all current references for unnecessary absolute path
- Setup a GitHub action to check for these anti-patterns (we can also check if the script is properly set to executable as well)

PR #1689, opened 2021-09-05
Remove unnecessary `/usr/sbin` and `/sbin` references

kidonng:
See #1688

A [remaining reference](https://github.com/matryer/xbar-plugins/blob/b74b4799e6020df9e2d88c6ef536e852e167840c/Network/updatevpnbookpass.1d.sh#L73) is in `Network/updatevpnbookpass.1d.sh` as it also references a path `/Applications/Utilities/Keychain Access.app` which no longer existing on newer systems, so I guess that line is broken (hence commented).

### DIFF
--- a/Dev/Kubernetes/minikube.5s.sh
+++ b/Dev/Kubernetes/minikube.5s.sh
@@ -13,6 +13,7 @@ MINIKUBE_LOGO_BW='iVBORw0KGgoAAAANSUhEUgAAAIAAAAB8CAYAAAChbripAAAABGdBTUEAALGPC/
 
 LANG="en_US.UTF-8"
 # You may need to update the PATH to include minikube and kubectl
+#PATH="/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
 export LANG PATH
 
 if [[ $(minikube status | head -n 1) =~ Running ]]; then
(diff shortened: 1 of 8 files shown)


Dev/Kubernetes/minikube.5s.sh
```diff
@@ -13,7 +13,6 @@ MINIKUBE_LOGO_BW='iVBORw0KGgoAAAANSUhEUgAAAIAAAAB8CAYAAAChbripAAAABGdBTUEAALGPC/
 
 LANG="en_US.UTF-8"
 # You may need to update the PATH to include minikube and kubectl
-#PATH="/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
```
kidonng, line 16: This line only prepends `/usr/local/sbin` to the default path but `minikube` or `kubectl` won't   be there, so this line essentially does nothing.

hagemt:
+1 for this; I noticed this is a problem in all three Homebrew dev plugins, too.

Dev/Kubernetes/minikube.5s.sh
kdeldycke: For reference, this is the original commit for `meta_package_manager.7h.py`: https://github.com/kdeldycke/meta-package-manager/commit/f35abda586d4275f730c5a9e74805a8a1eb6d35a

matryer:
thanks @kidonng

PR #1708, opened 2021-10-07
Find all the tags instead of '[firsttag', as currently.

pradzins:
The problem might have been caused by a change made to the output
of watson status command. In any case, this fix makes it compatible
with recent Watson versions.


sprak3000 (approved):

pradzins:
Thank you, @sprak3000  :-)

PR #1709, opened 2021-10-08
My quick-n-dirty new mail counter

albinoloverats:
A simple Python plugin that checks any number of email accounts via IMAP and displays the number of new/unread messages.


sprak3000:
@albinoloverats 

Please add a `<xbar.image>` metadata tag linking to a screenshot of the plugin in action.

PR #1715, opened 2021-11-01
feat: new brew-updates plugin which also replaces homebrewcask plugin

jimeh:
New brew-updates plugin which supports:

- Showing outdated formulas, casks, and pinned formulas, with details
  about exact version numbers.
- Upgrading:
  - Individual formulas and casks.
  - All formulas.
  - All casks.
  - All formulas and casks.
- Pinning and unpinning formulas.
- Uninstalling formulas, casks, and pinned formulas.

This can safely replace the existing homebrewcask plugin, especially as
it seems to have been broken ever since Homebrew removed the "cask"
sub-command.

It can also replace the existing brew-updates plugin, which is very
basic in that it simply lists all outdated formulas and casks, with the
only action available being to upgrade, without showing details about
current/latest versions.

---

I'm not sure what the policy is on removing existing plugins, so apologies if simply removing them in a PR like this is not the right way to go about it. I'm happy to change anything as needed :)

### Screenshots

New brew-updates:

new-brew-updates

Old brew-updates (removed):

Screen-Shot-2021-11-01-00-47-46 70

homebrewcask (removed):

homebrewcask


jimeh:
Apologies for the force pushes. I've simply been updating my PR's here to include updates/fixes I've been making in my [dotfiles](https://github.com/jimeh/dotfiles/tree/main/xbar).

That said, I do not expect to make any further changes for some time, as I've now resolved all annoyances I've had with this script.

sprak3000:
@jimeh 

Looks like the branch has a conflict. Could you clear that up?

jimeh:
@sprak3000 all done, rebased on top of the lastest state in `main`, and also updated the script with a minor update from my dotfiles.

sprak3000 (approved):

PR #1717, opened 2021-11-01
feat: rewrite brew-services plugin

jimeh:
This is a complete rewrite of the brew-services plugin, based on my
brew-updates plugin.

Main functional differences from the original plugin is:

- The option to group started/stopped services together.
- Hide services you don't want shown on the main list.
- Visual style being a bit different.
- An option to uninstall stopped services.
- Overall it doesn't suffer from any of the bugs that I'd run into with
  the old one.

---

I'm not sure what the policy is with a wholesale plugin rewrite/replacement like this, so I'm happy to make any changes as needed :)

### Screenshots

New brew-services:

new-brew-services
Screen-Shot-2021-11-03-14-05-42 13

Old brew-services:

old-brew-services

### DIFF
--- a/Dev/Homebrew/brew-services.10m.rb
+++ b/Dev/Homebrew/brew-services.10m.rb
@@ -1,119 +1,534 @@
 #!/usr/bin/env ruby
+# frozen_string_literal: true
 
 # <xbar.title>Brew Services</xbar.title>
-# <xbar.version>v1.0</xbar.version>
-# <xbar.author>Adam Lindberg</xbar.author>
-# <xbar.author.github>eproxus</xbar.author.github>
-# <xbar.desc>Shows and manages Homebrew services.</xbar.desc>
-# <xbar.image>http://i.imgur.com/hVfhHYP.jpg</xbar.image>
-# <xbar.dependencies>ruby, brew, brew-services</xbar.dependencies>
+# <xbar.version>v3.0.1</xbar.version>
+# <xbar.author>Jim Myhrberg</xbar.author>
+# <xbar.author.github>jimeh</xbar.author.github>
+# <xbar.desc>List and manage Homebrew Services</xbar.desc>
+# <xbar.image>https://i.imgur.com/gIQki4q.png</xbar.image>
+# <xbar.dependencies>ruby</xbar.dependencies>
+# <xbar.abouturl>https://github.com/jimeh/dotfiles/tree/main/xbar</xbar.abouturl>
+#
+# <xbar.var>boolean(VAR_GROUPS=true): List services in started/stopped groups?</xbar.var>
+# <xbar.var>string(VAR_BREW_PATH="/usr/local/bin/brew"): Path to "brew" executable.</xbar.var>
+# <xbar.var>string(VAR_HIDDEN_SERVICES=""): Comma-separated list of services to hide.</xbar.var>
 
-# BitBar Homebrew services plugin
-# by Adam Lindbeng (@eproxus)
+# rubocop:disable Lint/ShadowingOuterLocalVariable
+# rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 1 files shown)


jimeh:
Apologies for the force pushes, I came across a couple of bugs related to some edge cases and issues with brew services itself. These have now been addressed.

eproxus:
🙌 I wrote the original plug-in and wholeheartedly endorse this rewrite.

jimeh:
Apologies for the force pushes. I've simply been updating my PR's here to include updates/fixes I've been making in my [dotfiles](https://github.com/jimeh/dotfiles/tree/main/xbar).

That said, I do not expect to make any further changes for some time, as I've now resolved all annoyances I've had with this script.

jimeh:
Latest force push brought in the latest fixes from my dotfiles repo to add support for the new `scheduled` and `none` service statuses, and also switches from parsing the plain text output of `brew services list`, to parsing the JSON output when passing the `--json` flag. This should ensure the script is less likely to break due to future changes to brew services.

martin-braun:
It does not update the state when a service is started/stopped. Perhaps it re-runs the script instantly, but brew takes a little while until changes are registered.

@jimeh can anything be done about it?

martin-braun:
@jimeh may I make a feature request before this PR gets accepted? It would be nice if I could hide some services in a sub-menu, because they are getting installed by other packages for their libs, but I don't use their services. 

I.e. I install php and get a service along, but I only use php from the command line. I really wish those to be hidden in a sub menu and not counted as well. 

A blacklist config would be sufficient, but a way to hide and unhide entries via the context menu would be superior.

jimeh:
@martin-braun thanks for the feedback, I'll see what I can do adding some state to the plugin to enable a hide/unhide menu item :)

And you are right, `3/14` in the menu bar does look like 3 out of 14 are running, rather than `3 running / 14 stopped` as it reads when you open the menu. I'll probably just change it to show only the running count, or maybe make it configurable, we'll see :)

jimeh:
@martin-braun I've updated the plugin. Main number display now only shows running services, and hide/show option is available on services, along with an option to enable/disable groups too while I was already adding support to save state :)

screenshot

martin-braun:
@jimeh This is awesome, thank you so much ❤️! I think it is much better with only one number to save some precious real-estate in the top bar. And finally, I can hide those services that I never need. Awesome work!

Now for pure perfect this bulb would turn off when no services would run, but the emoji table offers nothing. I'm happy nonetheless.

I hope your plugin will be merged eventually. Great contribution.

sprak3000:
@jimeh 

Tried running this plugin, but I don't have any services managed / installed by brew. This results in the plugin throwing an error. Any chance you could update the plugin to account for this behavior?

jimeh:
@sprak3000 Ah, damn. That's one edge case I hadn't encountered myself... lol

I'll put a fix together tomorrow when I'm back on a computer :)

jimeh:
@sprak3000 apologies for the delay, this week turned out busier than I expected. But I've pushed a new v3.0.1 which works correctly when `brew services list --json` returns nothing.

sprak3000 (approved):

sprak3000:
@jimeh 

No need for apologies. Thank you for the update. Merged!

onaforeignshore:
Another check for all Homebrew plugin authors. Homebrew now installs in /opt/homebrew/bin/brew on any ARM Macs by default. Only checking /usr/local/bin/brew is insufficient.

jimeh:
@onaforeignshore Thanks for the reminder.  I'll sort it a fix in the next day or two.

jimeh:
@onaforeignshore Brew path detection should be fixed in #1921, feel free to comment there if you have any issues :)

PR #1728, opened 2021-11-15
Add github notification token variables

fmenezes:
Screen Shot 2021-11-15 at 10 35 01

The image is self-explanatory, this adds new settings for the plugin.


fmenezes:
cc @flesch

PR #1744, opened 2021-12-13
feat: request to add guild price tracker

rririanto:
$Guild price tracker will help our token holder track our token price in different currencies.


matryer (approved):

0220tiezo1323 (commented):
No

PR #1745, opened 2021-12-16
remove dependencies and enhance the UX of the Air Quality Index plugin.

nicoster:
I'd like to improve the UI a little bit and then found out that the aqicn.org has some nice graphs and here is what it looks like now:

and it doesn't depend on the `jq`, since we're dealing with the HTML tags. 
TODO: add donation info from aqicn.org to support them.


nicoster:
Hi @matryer, anything I can do to improve the code so it can be merged to the repo?

sprak3000 (approved):

PR #1752, opened 2022-01-11
Add new plugin for github version of gitlab_merge_requests.js

SimeonC:
My company moved from gitlab to github so I re-wrote my plugin to work with github to keep the same workflow.

CleanShot 2022-01-11 at 12 23 41


SimeonC:
@matryer Have updated the file name to include the default timeout as in the docs (missed this earlier). Is there anything else blocking this from being merged?

viperfx:
@SimeonC how do you currently set this up? I get an error

```
import fs from "fs";
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

SimeonC:
Hmmm, it should just work due to the mjs file extension. What version of node.js are you running?

sprak3000 (approved):

PR #1767, opened 2022-01-27
Add iCloud Private Relay status indicator

arice:
This plugin checks if the IP used for an insecure http request can be found on Apple's list of egress IPs used by [iCloud Private Relay](https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay) and shows a happy green sparkle if so, or a scary red flag if not. This plugin is based on [My External IP](https://github.com/matryer/xbar-plugins/blob/main/Network/external-ip-1.0.1h.sh) by CodinCafe.


matryer (approved):
Great plugin.

PR #1791, opened 2022-03-08
Secure TOTP

micxer:
Other plugins keep generating tokens and put them into calls of the plugin menu. This only shows the names in the menu and generates the token only when clicking on an entry. After 30 seconds the clipboard is cleared again.
Also the secrets do not have to be inserted into the script but can be placed anywhere on disk and referenced as a file.
And since the tokens are only generated when an entry is clicked, this plugin only has to be refreshed, when there is a new entry in the secrets file.

### DIFF
--- a/System/secure-totp.sh
+++ b/System/secure-totp.sh
@@ -4,18 +4,26 @@
 # <xbar.version>v1.0</xbar.version>
 # <xbar.author>micxer</xbar.author>
 # <xbar.author.github>micxer</xbar.author.github>
-# <xbar.desc>When you click on an entry, the token gets copied to the clipboard. After 30 seconds, the clipboard is automatically emptied again. The token are not part of the menu all the time.</xbar.desc>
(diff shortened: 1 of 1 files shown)


System/secure-totp.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+
+# <xbar.title>Secure TOTP Authenticator</xbar.title>
+# <xbar.version>v1.0</xbar.version>
+# <xbar.author>micxer</xbar.author>
+# <xbar.author.github>micxer</xbar.author.github>
+# <xbar.desc>When you click on an entry, the token gets copied to the clipboard. After 30 seconds, the clipboard is automatically emptied again. The token are not part of the menu all the time.</xbar.desc>
+# <xbar.image>https://i.imgur.com/FelKuk7.png</xbar.image>
```
matryer, line 8: Hey @micxer - please use an image of the plugin in use.

sprak3000:
@micxer 

Poking again regarding the `<xbar.image>`. Could you change it to be a screenshot of the plugin in use?

micxer:
Good things come to those who wait. 😄

sprak3000 (approved):

PR #1810, opened 2022-04-04
New plugging to check unread emails from any IMAP mailbox

maesitos:
Very simple plugin to check unread email from any IMAP server

No new email:
Screenshot 2022-04-04 at 20 12 09

Unread email:
Screenshot 2022-04-04 at 20 10 24

In case there is any email and the webmail is configured it'll show a direct link:

Screenshot 2022-04-04 at 20 30 53 Screenshot 2022-04-04 at 20 28 35


sprak3000:
@maesitos 

sprak3000 (approved):

PR #1815, opened 2022-04-14
Add new plugin: Mastercard currency tracker script

alpsayin:
I've noticed many of the currency trackers don't work, or they require python2, which I refuse to install any more 🥲 .
So here's a non-api-key-required currency rate ticker for a single pair.

Plugin image

Edit: used the mastercard script as a basis and forked it out to 3 different sripts utilising google and duckduckgo

##### Google;
xbar_currency_tracker_duckduckgo

##### Duckduckgo;
xbar_currency_tracker_duckduckgo


alpsayin:
Decided to publish in my own repository :)

PR #1828, opened 2022-05-14
Add TypeScript as Supported languages in CONTRIBUTING.md

theogainey:
This pr adds documentation that TypeScript and JavaScript plugins can be written using [Deno](https://deno.land/) as a runtime. Deno treats TypeScript as a first class language meaning that TypeScript does not need to be transpiled to JavaScript.


Weather/weather-alerts.1min.ts
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env -S -P/${HOME}/.deno/bin:/opt/homebrew/bin deno run --allow-net
```
sprak3000, line 1: @theogainey 

Would it make sense to add `/usr/local/bin` (the non-M1 silicon Homebrew path) here?

sprak3000 (changes requested):
@theogainey 

Please run `chmod +x` on your script file. I also left a comment regarding the path to `deno`.

Weather/weather-alerts.1min.ts
theogainey: Yes it would so I added it

theogainey:
I run chmod +x on the script file

sprak3000 (approved):

PR #1831, opened 2022-05-22
Make terminal not show and refresh bar

Daniel-Fernandez-951:
Rearranging order acts as intended:
- Terminal will NOT show when a new DNS profile selected
- Bar WILL refresh when a new DNS profile is selected


Daniel-Fernandez-951:
> Rearranging order acts as intended:
> 
> * Terminal will NOT show when a new DNS profile selected
> * Bar WILL refresh when a new DNS profile is selected

Tested on SwiftBar 1.4.3

sprak3000 (approved):

PR #1848, opened 2022-07-30
Update SureFlap plugin for xbar

henrik:
- Use variables for config
- Customisable in/out display
- Shorter error output, as overlong output would cause empty output (might depend on notch etc)


sprak3000:
@henrik 

henrik:
Done! Cheers.

sprak3000 (approved):

PR #1869, opened 2022-10-20
Add Github Release buddy

luis-santos-teampicnic:
Adds a new plugin that shows the latest release in Github in a specific set of repositories.
Configuration is done through the xbar variables system.

Example output:
Screenshot 2022-10-23 at 17 18 31

Configuration is done through the xbar variables system:
Screenshot 2022-10-23 at 17 25 11


Dev/GitHub/github-releases.30m.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# <xbar.title>GitHub Relase buddy</bitbar.title>
```
matryer, line 4: nice idea - please provide an example picture of the plugin running.

luis-santos-teampicnic: Added the image to the PR and to the plugin metadata! :)

sprak3000 (approved):

PR #1890, opened 2023-01-08
Remove broken or obsolete plug-ins

dnicolson:
This pull request removes mostly Python 2 plug-ins that no longer work due to APIs no longer being available or the company no longer existing.

Fixes https://github.com/matryer/xbar-plugins/issues/1564, https://github.com/matryer/xbar-plugins/issues/1530.


Science/iss-overhead-check.60m.py
```diff
@@ -1,47 +0,0 @@
...
-from time import localtime
-from requests import get
-
-LATITUDE = -27.591665  # Put your latitude here
-LONGITUDE = -48.589599  # Put your longitude here
-ALTITUDE = 12  # Put your altitude here
-
-
-def iss_overhead():
-    url = "http://api.open-notify.org/iss-pass.json?lat={}&lon={}&alt={}".format(LATITUDE, LONGITUDE, ALTITUDE)
```
sprak3000, line 20: Looks like this API is now at http://api.open-notify.org/iss-now.json. Any chance of salvaging this plugin using it?

Finance/currency-tracker.1h.py
```diff
@@ -1,38 +0,0 @@
...
-# Base comparaison currency
-currFrom = "EUR"
-# Array of tracked currencies
-currTo = ["CAD", "USD"]
-
-urlParamTo = currTo[0]
-if len(currTo) > 1:
-    urlParamTo = ",".join(currTo)
-
-url = "https://api.exchangeratesapi.io/latest?base=" + currFrom + "&symbols=" + urlParamTo
```
sprak3000, line 24: This still seems to be a valid API. What is the reason behind removing this plugin?

Time/shiftstats.30s.py
```diff
@@ -1,143 +0,0 @@
-#!/usr/bin/python
```
sprak3000, line 1: What is broken or obsolete in this plug-in? Any chance of salvaging it?

Tools/Slack/slack-unread.1s.py
```diff
@@ -1,209 +0,0 @@
...
-slack_call_command = ' href=slack:'
-
-channels_list_url = 'https://slack.com/api/channels.list'
```
sprak3000, line 46: Guessing the Slack API has changed. Is there another up-to-date plugin that fills this need? If not, any chance of salvaging this?

Web/faith_in_humanity_score.6h.py
```diff
@@ -1,16 +0,0 @@
...
-# <xbar.version>v1.0</xbar.version>
-# <xbar.author>impshum</xbar.author>
-# <xbar.author.github>impshum</xbar.author.github>
-# <xbar.desc>Displays current score from https://faithinhumanityscore.com</xbar.desc>
-#
-# by impshum
-
-from requests import get
-
-data = get('https://faithinhumanityscore.com/api.php').json()
```
sprak3000, line 14: Looks like there might be a way to salvage this plugin.

@impshum Any interest in updating this?

sprak3000 (commented):
Seeing if there is any chance of salvaging a few of these, if there are already alternatives to them, etc.

Web/faith_in_humanity_score.6h.py
impshum: Fixed it: https://faithinhumanityscore.com/api.php

Science/iss-overhead-check.60m.py
dnicolson: I don't think so, it looks like the whole API would need to be implemented from https://github.com/open-notify/Open-Notify-API/blob/88c34b793c9b0b299c8e941fd477e12c0b974353/app.py#L108.

Finance/currency-tracker.1h.py
dnicolson: The API is no longer open and the free tier only allows for 250 requests per month.

Time/shiftstats.30s.py
dnicolson: It depends on `/var/log/accountpolicy.log` which seems to unavailable for more than 6 years.

Tools/Slack/slack-unread.1s.py
dnicolson: I tested this with an App-Level-Token but was getting `invalid_auth` after replacing the [deprecated methods](https://api.slack.com/changelog/2020-01-deprecating-antecedents-to-the-conversations-api), I will restore this as it may still be possible somehow to make it work.

Science/iss-overhead-check.60m.py
sprak3000: @mrprompt 

Any interest in updating this plugin?

Finance/currency-tracker.1h.py
sprak3000: 🤔 I'm not sure we want to remove this entirely. Free tier would still allow you to track things, albeit not on an hourly schedule.

sprak3000:
@dnicolson 

We likely have enough currency trackers in some form to cover the loss of that plugin. Let's give the author of the ISS plugin a chance to respond and see if they want to overhaul it. If they don't reply within a week, I'll merge this.

impshum:
@dnicolson

> We likely have enough currency trackers in some form to cover the loss of that plugin. 

This is not a currency tracker. It's for my website: https://faithinhumanityscore.com

PR #1937, opened 2023-04-05
Add Zoom Audio/Video/Share Status plugin

romansavrulin:
Creates a nifty zoom audio/video/share status icons in status bar, so you can always see if you are on air or not

## On Air

Mic is activated
image

Camera is activated
image

Screen share is activated
image

Any combinations of the previous states are supported

## In conference, no sources are activated
image

## Not in conference
image

## Details

Script is very simple, but supports Zoom localization, so ought to work on different machines

It requires Acessibility permissions to be allowed to the swiftbar


sprak3000:
@romansavrulin 

1. The metadata tags need to be `<xbar>` not `<bitbar>`.
2. Need to run `chmod +x` on the script file.

romansavrulin:
Thanks. I've checked this script with swiftbar. Will need to adapt to xbar and test it once again. Will notify you when ready

romansavrulin:
Had to change icons and a logic a bit, due to lack of support of sfcolors/sf icons
image

sprak3000 (approved):

PR #1946, opened 2023-04-13
Update weather to get location from IP

MilapNaik:
For me this is great since I travel a lot. I think it will help others.


MilapNaik:
@sprak3000 Yes I've encountered the VPN situation too, would love a fix for that. I can take a look at using that.

sprak3000 (approved):

PR #1949, opened 2023-04-30
feat: ram load plugin

gustavoilhamorais:
Adds new RAM (random access memory) plugin to display RAM load on OSX topbar under XBAR management.

Default refresh rate is set to five seconds.

Plugin is written in Python3 and allow user to choose between six display options: 
* bytes
* KiloBytes
* MegaBytes
* GigaBytes
* Available percentage
* Used percentage (default)

Python3 dependencies:
* psutil

Extra features:
* Captures 'ModuleNotFoundError' exception and automatically insert into clipboard the commad to install the dependencies under correct Python environment. Renders an error message instead of expected data and extra instructions on dropdown.


System/ram-load.5s.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python3
```
sprak3000, line 1: This is the shebang most plugins have begun adopting to ensure people using python3 from OS X, installed from homebrew on Intel chip, and installed from homebrew on M1 chip can run it without modifying the script.

```suggestion
#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
```

sprak3000 (changes requested):
@gustavoilhamorais 

Works great! Just need to run `chmod +x` on the script, and we could merge it. I've left one other comment on the shebang line to help compatibility across various machines. Not necessary to address for merge though.

System/ram-load.5s.py
gustavoilhamorais: Have updated the shebang as proposed. Thanks for the tip.

sprak3000 (approved):

PR #1959, opened 2023-06-06
Final Commit

AndrewJ41:
Tidies up the code a little - makes font size configurable and adds in current usage from the inverter and from the grid. This is about as far as this plugin can go for now without adding in EV Charging. 

NOTE: Will revisit when I have an electric car - already have the inverter.

### DIFF
--- a/Environment/solaredge.5m.py
+++ b/Environment/solaredge.5m.py
@@ -1,15 +1,16 @@
 #!/usr/bin/python3
 # -*- coding: utf-8 -*-
 # <xbar.title>SolarEdge Monitoring</xbar.title>
-# <xbar.version>v1.1</xbar.version>
+# <xbar.version>v2.0</xbar.version>
 # <xbar.author>Shaun Grady</xbar.author>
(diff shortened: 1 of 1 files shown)


Environment/solaredge.5m.py
```diff
@@ -19,12 +20,12 @@
...
 solaredge_site_id = os.getenv("SITE_ID", "")
 solaredge_api_key = os.getenv("API_KEY", "")
+font_size = os.getenv("FONT_SIZE", "13")
 
 # Optional. Set to 0 to disable. Total panel DC watt capacity
-system_wattage = 0
+system_wattage = 5460
 
 # Optional. Set to 0 to disable. Find CO2 lbs/MWh for your subregion in the PDF linked below.
 # (Use the "Non-baseload output emission rates" figure for your subregion)
```
sprak3000, line 31: Should these optional items become variables?

sprak3000 (commented):
Looks good. Just one question around some of the optional items seen in the code.

PR #1960, opened 2023-06-09
Use SolarEdge Environmental Figures

AndrewJ41:
- Fixed a bug where battery charge power was getting added into total DC Output, this should only be the case if the battery is in a discharge state
- Removed Environmental Impact calculation and replaced with SolarEdge stats
- Added indicator to show if battery is in charge, discharge or idle state.
- Made all optional or configurable elements xbar variables.

### DIFF
--- a/Environment/solaredge.5m.py
+++ b/Environment/solaredge.5m.py
@@ -12,7 +12,7 @@
 # <xbar.var>string(BATTERY_PRESENT): Does your SolarEdge installation include a battery? Y/N</xbar.var>
 # <xbar.var>string(SYSTEM_WATTAGE): Total panel DC watt capacity</xbar.var>
 # <xbar.var>string(SHOW_ENVIRONMENTAL_BENEFIT): Do you want to show your environmental benfit in the dropdown? Y/N (Default Y)</xbar.var>
-# <xbar.var>string(ENVIRONMENTAL_BENEFIT_UNIT): Units to display benefit in? Imperial/Metrics (Default Metrics)</xbar.var>
+# <xbar.var>select(ENVIRONMENTAL_BENEFIT_UNIT="Metrics"): Units to display benefit in? [Imperial, Metrics]</xbar.var>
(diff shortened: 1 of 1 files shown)


Environment/solaredge.5m.py
```diff
@@ -10,6 +10,9 @@
 # <xbar.var>string(SITE_ID): Please provide your SolarEdge installation SiteId</xbar.var>
 # <xbar.var>string(API_KEY): Please provide your SolarEdge installation API Key (https://www.solaredge.com/node/88689)</xbar.var>
 # <xbar.var>string(BATTERY_PRESENT): Does your SolarEdge installation include a battery? Y/N</xbar.var>
+# <xbar.var>string(SYSTEM_WATTAGE): Total panel DC watt capacity</xbar.var>
+# <xbar.var>string(SHOW_ENVIRONMENTAL_BENEFIT): Do you want to show your environmental benfit in the dropdown? Y/N (Default Y)</xbar.var>
```
sprak3000, line 14: Why not a boolean variable here?

Environment/solaredge.5m.py
```diff
@@ -10,6 +10,9 @@
...
+# <xbar.var>string(ENVIRONMENTAL_BENEFIT_UNIT): Units to display benefit in? Imperial/Metrics (Default Metrics)</xbar.var>
```
sprak3000, line 15:
```suggestion
#  <xbar.var>select(ENVIRONMENTAL_BENEFIT_UNIT="Metrics"): Which style to use. [Imperial, Metrics]</xbar.var>
```

sprak3000 (commented):
@AndrewJ41 

Looks fine; just some suggestions around the variables.

Environment/solaredge.5m.py
AndrewJ41: More readability than anything else - can change to bool if you think it helps.

sprak3000: 🤷🏽 I'm not a python expert; if you feel this is more readable, that works for me.

PR #1963, opened 2023-06-15
Adding internet-status.sh

afrazkhan:
`internet-status.sh` is like `got-internet.sh`, but includes a check for DNS, with separate icons for ping, and DNS failures / success.

### DIFF
--- a/Network/internet-status.2s.sh
+++ b/Network/internet-status.2s.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# <xbar.title>Internet Status</xbar.title>
+# <xbar.version>v1.0</xbar.version>
+# <xbar.author>Afraz Ahmadzadeh</xbar.author>
+# <xbar.author.github>afrazkhan</xbar.author.github>
+# <xbar.desc>Checks network and DNS functionality</xbar.desc>
+
+ping_timeout=1
+ping_address=9.9.9.9
+dns_address=www.quad9.net
+
+# By default, everything is fine
+icon="🌎|dropdown=false"
+message="All is well"
+dns_icon="✅"
+ping_icon="✅"
+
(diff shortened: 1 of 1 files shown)


Network/internet-status.2s.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+# <xbar.title>Internet Status</xbar.title>
+# <xbar.version>v1.0</xbar.version>
+# <xbar.author>Afraz Ahmadzadeh</xbar.author>
+# <xbar.author.github>afrazkhan</xbar.author.github>
+# <xbar.desc>Checks network and DNS functionality</xbar.desc>
+
+ping_timeout=1
+ping_address=9.9.9.9
+dns_address=www.quad9.net
```
sprak3000, line 10: Not required to get this merged in, but something to consider. You could make these [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html) and allow users to set things based on their preferences.

sprak3000 (changes requested):
@afrazkhan 

Looks good but needs two small updates:

1. Run `chmod +x` on the file.
2. Add an `<xbar.image>` metadata tag linking to a screenshot of the plugin in action. What does it look like in the bar with the menu open?

Network/internet-status.2s.sh
matryer: @afrazkhan what do you think of that? Could be worth it.

sprak3000 (approved):
Added in the image and variables use. Will `chmod` in separate PR once merged.

Network/internet-status.2s.sh
afrazkhan: Sorry, I missed this, but thanks for taking care of it for me :)

PR #1966, opened 2023-06-18
Add GitHub PRNotii Plugin

eu81273:
This is a plugin that tells GitHub Pull Request status case by case.
As you know, there are many Pull Request notification plugins, but it was difficult to find a plugin suitable for actual team work.
This plugin can be viewed by type and situation of Pull Request, such as PR that needs review, PR that has already been reviewed, and PR that I created.
This plugin has a very well-organized code and is easier to use.

Thank you.

PRNotii


Dev/GitHub/prnotii.5m.js
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env node
+
+// <xbar.title>GitHub PRNotii</xbar.title>
+// <xbar.version>v1.0</xbar.version>
+// <xbar.author>aiden.ahn</xbar.author>
+// <xbar.author.github>eu81273</xbar.author.github>
+// <xbar.desc>Check Github Pull Reuqest</xbar.desc>
+// <xbar.image>https://github.com/matryer/xbar-plugins/assets/2687916/e08d6380-62f9-478b-b35b-e2222cf65791</xbar.image>
```
sprak3000, line 8: Should use the actual image path here.

```suggestion
// <xbar.image>https://github-production-user-asset-6210df.s3.amazonaws.com/2687916/246659270-e08d6380-62f9-478b-b35b-e2222cf65791.png</xbar.image>
```

Dev/GitHub/prnotii.5m.js
```diff
@@ -0,0 +1,108 @@
...
+      case e.message.includes('ENOENT'):
+        printOut('🌈 HOW TO SETUP');
+        printOut('---');
+        printOut('❶ Generate Github Token');
+        printOut('⠀  Generate new personal access token with "repos scope". | size=12');
+        printOut('❷ Create .prnotii File');
+        printOut('⠀  Create ~/.prnotii File like below JSON format. | size=12');
+        printOut('⠀  {"token": "kSEzyDP...", "endpoint": "https://api.github.com"}  | size=12');
+        printOut('❸ Press Refresh button below');
+        printOut('⠀  After finish ❶,❷ steps, press "Refresh" button to start again. | size=12');
```
sprak3000, line 91: This looks like a good use case for [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html). User would fill in the variables through the plugin browser UI, and the JSON configuration file is written automatically. You would just need to pull in the values via environment variables after that.

Not sure though if SwiftBar has a similar mechanism. If it doesn't, might be able to have the best of both systems with this flow:

1. Check for environment variables. If present, use them.
2. If no environment variables present, try reading the `.prnotii` configuration file.
3. If `.prnotii` not found, then display setup instructions. If able to detect we are running xbar versus swiftbar (I believe both have an env variable you could use for this purpose), write the setup instructions out for the appropriate program.

What are your thoughts?

sprak3000 (changes requested):

Dev/GitHub/prnotii.5m.js
eu81273: @sprak3000 
Thank you for the suggestion. I was able to learn about the xbar variables feature thanks to you. It seems like a perfect feature for this case. I have changed the implementation to handle the xbar variable file directly in order to minimize changes to the existing code.

sprak3000: @eu81273 

I'm trying to test this locally and am getting `Unexpected token R in JSON at position 2`. I went through the xbar plugin browser UI to have it create the JSON configuration file. It looks like this:

```json
{
        "VAR_ENDPOINT": "https://api.github.com",
        "VAR_TOKEN": "ghp_..."
}
```

Any ideas on why the `JSON.parse` seems to be failing?

Dev/GitHub/prnotii.5m.js
```diff
@@ -0,0 +1,113 @@
...
+  href: item.html_url,
+  title: item.title,
+}));
+const organize = refined => refined.reduce((prev, item) => {
+  prev[item.repository] = prev[item.repository] || [];
+  prev[item.repository].push(item);
+  return prev;
+}, {});
+const githubQuery = (endpoint, token, query) => 
+  request(endpoint + '/search/issues?q=' + encodeURIComponent(query), {'Authorization': 'token ' + token})
```
sprak3000, line 45: @eu81273 

Reading in the configuration is actually working fine. The error I'm encountering (`Unexpected token R in JSON at position 2`) is this line. The GitHub API requires a user agent sent in the headers. This cleared things up for me:

```suggestion
  request(endpoint + '/search/issues?q=' + encodeURIComponent(query), {'User-Agent': 'xbar-app', 'Authorization': 'token ' + token})
```

I'll leave it to you to decide what you might want to put in as the user agent. You could promote your work in -- `prnotti-for-xbar-by-eu81273`. 😄

sprak3000 (changes requested):

Dev/GitHub/prnotii.5m.js
sprak3000: The above was not the issue. Turns out not having a user agent in the request was. See latest review below.

sprak3000 (approved):
Really nice work here; thank you for your submission!

Issue #1968, opened 2023-06-20
001-emojiWeather.30m.js: request module has been deprecated


Ischca:
fao @Chandler-Davidson - 

The request module is obsolete.
No need to respond right away.
If I have time I will create a PR at a later date.

https://www.npmjs.com/package/request

Chandler-Davidson:
@sprak3000, thanks for the contribution!

PR #1969, opened 2023-06-20
Issue #1968 - 001-emojiWeather.30m.js: request module has been deprecated

sprak3000:
- Replace `request` with `axios` library.
- Fix emoji generation. Based on the documentation, the weather ID to emoji calculation and list of emojis itself was not in sync.
- Move city and API key to xbar variables.
- Update shebang to use `env` pattern.


sprak3000:
@Ischca

Would you be able to give this a test and see if it works for you?

Ischca:
Good!
LGTM and thanks for the many fixes.

PR #1970, opened 2023-06-21
Changed template images

eu81273:
I aware that starting with macOS Big Sur, the color of system tray icon automatically changes based on the brightness of the background screen.

With templateImage parameter, the color of the icon image is automatically converted, so I changed to use templateImage instead of image.

However, if tint color can be set for templateImage, it seems that color can be expressed as in the image below, but xbar does not seem to have this option. Is that right?

스크린샷 2023-06-20 오후 11 50 54


sprak3000 (approved):
I can't speak to the tint question; @matryer might have more insight on that. Otherwise, this looks good.

PR #1972, opened 2023-06-22
Added nightshift toggler

andersonaddo:
A quick xbar tool to toggle MacOS Nightshift with less clicks.

### DIFF
--- a/System/macosnightshift.1h.sh
+++ b/System/macosnightshift.1h.sh
@@ -1,11 +1,13 @@
-#!/usr/bin/env bash
+#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" bash 
+#^added homebrew install locations (both pre-M1 and M1) to PATH
 
-#  <xbar.title>Mac Low Power Mode Manager</xbar.title>
+#  <xbar.title>Night Shift Toggler</xbar.title>
 #  <xbar.version>v1.0</xbar.version>
 #  <xbar.author>Anderson</xbar.author>
 #  <xbar.author.github>andersonaddo</xbar.author.github>
-#  <xbar.desc>A quick xbar tool to toggle MacOS Nightshift with less clicks. Check out https://github.com/andersonaddo/xbar-macos-nightshift-toggler for important installation instructions!</xbar.desc>
-#  <xbar.image>https://raw.githubusercontent.com/andersonaddo/xbar-macos-nightshift-toggler/main/icon.png</xbar.image>
+#  <xbar.desc>A quick xbar tool to toggle MacOS Night Shift with less clicks. Check out https://github.com/andersonaddo/xbar-macos-nightshift-toggler for important installation instructions!</xbar.desc>
+#  <xbar.dependencies>nightlight (Homebrew)</xbar.dependencies>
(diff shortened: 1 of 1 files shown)


System/macosnightshift.1h.sh
```diff
@@ -0,0 +1,64 @@
...
+
+#Nightshift modes
+OFF=0
+ON=1
+NOT_SUPPORTED=2
+
+nightshift_mode=$OFF
+
+get_nighshift_state() {
+  local nightshift_mode_raw=$(/usr/local/bin/nightlight status)
```
sprak3000, line 23: Homebrew installs software to a different path on M1 Macs: `/opt/homebrew/bin`. You can just do

```suggestion
  local nightshift_mode_raw=$(nightlight status)
```

if you use the `-S PATH=` suggestion above.

System/macosnightshift.1h.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
sprak3000, line 1:
```suggestion
#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" bash
```

This will allow tools installed by Homebrew to be used without specifying the full path. Homebrew on M1 Macs install into a different directory. This will catch both of them.

System/macosnightshift.1h.sh
```diff
@@ -0,0 +1,64 @@
...
+    nightshift_mode=$ON
+  elif [[ $nightshift_mode_raw == *"off"* ]]; then
+    nightshift_mode=$OFF
+  else
+    nightshift_mode=$NOT_SUPPORTED
+  fi
+}
+
+toggle_nightshift_state() {
+  /usr/local/bin/nightlight toggle
```
sprak3000, line 35: Again, you can just do

```suggestion
  nightlight toggle
```

if you use the `-S PATH=` suggestion above.

sprak3000 (changes requested):
Along with the suggestions below, you need to run `chmod +x` on your script file.

andersonaddo:
Thanks for all the suggestions, I'll get those done.
Not sure about the chmod one though: I I've already done that on my machine, but isn't that a something that has to be done to the on the machine of the user? I was under the impression that xbar does that, and that we were only being told to do that when testing our own scripts.

https://github.com/matryer/xbar-plugins/pull/1888 got merged last time without this comment so I'm genuinely curious.
(Btw, how some I still can't see #1888 on the Plugin directory?)

sprak3000:
> Thanks for all the suggestions, I'll get those done. Not sure about the chmod one though: I I've already done that on my machine, but isn't that a something that has to be done to the on the machine of the user?

I could be done on the machine of the user. However, checking the file in on GitHub with the executable bit set means they don't have to. xbar (to the best of my knowledge) will not set the executable bit.

> #1888 got merged last time without this comment so I'm genuinely curious.

Oversight on my part. I'll have to fix that. Pobody's nerfect. 😄 

> (Btw, how some I still can't see #1888 on the Plugin directory?)

The plugin directory requires a site rebuild, something I can't do currently. I'll poke @matryer about getting that done.

andersonaddo:
Ah! I didn't know the permission bits are also checked in; I thought those are purely on a by-machine basis. Got it, thanks!

andersonaddo:
Fixed

sprak3000 (approved):

PR #1993, opened 2023-10-08
Add WorldTimeGrid plugin

mrc:
Here's a new plugin to show a grid of times for different time zones. Let me know if you have any notes or suggestions!

It shows the UTC time on the menu bar like this:
image

When you click it it shows a grid of the previous and next 12 hours, highlighting the current hour:
image

There's a bunch of configuration options, including which time zones to display, how many hours to go forward and backwards, and the font. It requires Python 3.

### DIFF
--- a/Time/WorldTimeGrid.10s.py
+++ b/Time/WorldTimeGrid.10s.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
 
 #  <xbar.title>WorldTimeGrid</xbar.title>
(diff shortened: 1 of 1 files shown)


Time/WorldTimeGrid.10s.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
```
sprak3000, line 1: One suggestion... Might consider using this for the shebang:

```suggestion
#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
```

Allows the plugin to pick up python3 from either macOS install, Homebrew on M1, or Homebrew on Intel.

PR #2006, opened 2023-12-31
Live Tennis: Adapted to new API changes in ATP website

anupsabraham:


Sports/live_tennis.1m.py
```diff
@@ -1,111 +1,220 @@
...
+import getopt
+import os
 import json
-from urllib.request import Request, urlopen
+import http.client
+import gzip
 
-atpworldtour_base_url = "http://www.atpworldtour.com"
-inital_scores_url = atpworldtour_base_url + "/en/-/ajax/Scores/GetInitialScores"
+FAVORITE_PLAYERS_FILE = "/var/tmp/bitbar_tennis_favourites.data"
```
sprak3000, line 19: This might be a good use case for [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html). Let the user set where they want the file saved.

anupsabraham: Should it be completely configurable by the user or are you suggesting to use something like a dropdown options? 
The code should also verify if the path exists, right?

sprak3000 (approved):

PR #2025, opened 2024-02-16
sunrise-sunset: create the sun indicator

kiprasmel:
when i saw that xbar accepts dynamically generated images, i was very excited. here we go.

sun-full


kiprasmel:
cc @sprak3000 good to go

sprak3000 (approved):

PR #2035, opened 2024-05-08
kubeswitcher.1m.sh scan for KUBECONFIG

Raboo:
Search if there is a KUBECONFIG set and apply it to the script.

This will probably fix https://github.com/matryer/xbar/issues/877.


Dev/Kubernetes/kubeswitcher.1m.sh
```diff
@@ -11,6 +11,8 @@
 # KUBECONFIG=~/.kube/config # Uncomment and set this if you need a custom kube config path.
```
sprak3000, line 11: Overall, the changes look fine. While you are here though, it might be good to convert this over to a [xbar variable](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html).

Raboo: Hmm, I was just mimicking what most other Kubernetes desktop clients use. i.e. the already set environment variables.
Not sure which is best. Use this a configurable option like xbar variable and force people to have dual config for their KUBECONFIG's or just go with scanning for the KUBECONFIG in the most common places where it is configured..??

Raboo: I forgot to say I do the scanning in line 15.

Raboo: Can I have shell commands inside the default value of "xbar variables"? i.e $(somecommand)?

Raboo: It was something I forgot that was there from before. I think the scanning part should be enough, it should cover 99.99% of all use cases i am manguessing. I will remove the line.

Raboo: Ok done!

sprak3000 (approved):

PR #2038, opened 2024-05-12
Timetable v1.0

cabinz:
It's a menu bar plug-in to display the current event from a customizable timetable.
The user only needs to define the daily timetable as a `.csv` file and configure the path to the CSV file in the script. The plug-in will display the current event in real time on the menu bar. The pop-up window will show the whole timetable pre-defined in the CSV file.

sample

The csv file is required to be formatted as below:
csv

### DIFF
--- a/Time/timetable.1m.py
+++ b/Time/timetable.1m.py
@@ -12,7 +12,7 @@
 import datetime
 import csv
 from dataclasses import dataclass
-from typing import List
+from typing import List, Union
 
 # Each line in the format "Begin_Time,End_Time,Event_Name"
 #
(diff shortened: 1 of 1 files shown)


Time/timetable.1m.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import List, Union
+
+# Each line in the format "Begin_Time,End_Time,Event_Name"
+#
+# Example (timetable.csv):
+# 23:30,07:30,sleep
+# 07:30,08:30,breakfast
+# 08:30,10:00,gym
+# ...
+CSV_TIMETAB = "/Applications/SwiftBar/timetable/.timetable.csv"
```
sprak3000, line 24: This is a good use case for [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html).

sprak3000 (changes requested):

Time/timetable.1m.py
cabinz: Hi, I've already created a commit to use xbar variabls to make the CSV file path configurable.

Btw, I was the user of [SwiftBar](https://github.com/swiftbar/SwiftBar) and I initially developed this plug-in app for it. It's your change requests made me realize this repo is originally for another earlier open source project called xbar 😂. May I know is there any essential differences / relationship between these two projects?

Moreover, is there any way in xbar to redirect the local plugin folder from the default path `/Users/xxx/Library/Application Support/xbar/plugins` to a customized folder? (e.g. `.../xbar-plugins/Enabled`) I didn't find an obvious entry for this in xbar. Thanks!

sprak3000 (approved):

PR #2053, opened 2024-08-18
google chrome CPU usage

alexandregz:
Show CPU usage from Google Chrome; hardcoded, but you can use param APP to change this. Also you need to change ICON manually.

### DIFF
--- a/System/google_chrome_use_cpu.1m.sh
+++ b/System/google_chrome_use_cpu.1m.sh
@@ -3,7 +3,7 @@
 #  <xbar.version>v1.0</xbar.version>
 #  <xbar.author>Alexandre Espinosa Menor</xbar.author>
 #  <xbar.author.github>alexandregz</xbar.author.github>
-#  <xbar.image>https://raw.githubusercontent.com/alexandregz/xbar-plugins-alexandregz/main/screenshot_cpu_chrome.png</xbar.image>
+#  <xbar.image>https://imgur.com/G7D2XS9</xbar.image>
 #  <xbar.desc>CPU usage from Google Chrome</xbar.desc>
 #  <xbar.dependencies>bash</xbar.dependencies>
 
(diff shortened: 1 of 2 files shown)


System/google_chrome_use_cpu.1m.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+#  <xbar.title>CPU Google Chrome</xbar.title>
+#  <xbar.version>v1.0</xbar.version>
+#  <xbar.author>Alexandre Espinosa Menor</xbar.author>
+#  <xbar.author.github>alexandregz</xbar.author.github>
+#  <xbar.image>https://imgur.com/G7D2XS9</xbar.image>
+#  <xbar.desc>CPU usage from Google Chrome</xbar.desc>
+#  <xbar.dependencies>bash</xbar.dependencies>
+
+APP="Google Chrome.app"
```
sprak3000, line 10: You could use [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html) for this and the icon. Then, this could become a more generically named `app_use_cpu.1m.sh` plugin with Google Chrome as the default example app on startup.

System/google_chrome_use_cpu.1m.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+#  <xbar.title>CPU Google Chrome</xbar.title>
+#  <xbar.version>v1.0</xbar.version>
+#  <xbar.author>Alexandre Espinosa Menor</xbar.author>
+#  <xbar.author.github>alexandregz</xbar.author.github>
+#  <xbar.image>https://imgur.com/G7D2XS9</xbar.image>
```
sprak3000, line 6:
```suggestion
#  <xbar.image>https://imgur.com/G7D2XS9.png</xbar.image>
```

alexandregz: I changed screenshot to see Google Chrome cpu usage and other app, for example QEMU, and I renamed plugin.

Also I added CPU icon (generic, for other apps not Chrome) and I used xbar vars.

Thx for your feedback!

sprak3000 (approved):

PR #2061, opened 2024-09-25
Add API_ENDPOINT option to "GitHub Review Requests TS" plugin

alan-nf:
This plugin works great but has the GItHub API hard-coded. On GitHub Enterprise this API URL is different and needs to be configured. The GitHub PR Notii plugin already has this option, so I'm just adding a similar one to this plugin.

### DIFF
--- a/Dev/GitHub/github-review-requests-TS-5m.ts
+++ b/Dev/GitHub/github-review-requests-TS-5m.ts
@@ -54,7 +54,7 @@ class DataFetcher {
 
   async fetch({ token, username, filters }: Init): Promise<SearchResults> {
     const { data } = await axios.post(
-      this.apiEndpoint,
+      this.apiEndpoint ?? 'https://api.github.com/graphql',
       { query: this.buildQuery(filters, username) },
       {
         headers: {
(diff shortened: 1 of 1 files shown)


Dev/GitHub/github-review-requests-TS-5m.ts
```diff
@@ -12,13 +12,16 @@
 //  <xbar.var>string(USERNAME=""): Github User name.</xbar.var>
 //  <xbar.var>string(ORGANIZATION=""): Filter PRs by Organization name. (optional)</xbar.var>
 //  <xbar.var>string(WIP_FILTER=""): Title included in a Work in Progress PR to exclude from the list. (optional)</xbar.var>
+//  <xbar.var>string(API_ENDPOINT=""): URL to your GitHub GraphQL Endpoint. (optional)</xbar.var>
```
sprak3000, line 15: Rather than list it as optional, consider having the main GitHub URL as the default value.

```suggestion
//  <xbar.var>string(API_ENDPOINT="https://api.github.com/graphql"): URL to your GitHub GraphQL Endpoint.</xbar.var>
```

Dev/GitHub/github-review-requests-TS-5m.ts
```diff
@@ -51,7 +54,7 @@ class DataFetcher {
 
   async fetch({ token, username, filters }: Init): Promise<SearchResults> {
     const { data } = await axios.post(
-      `https://api.github.com/graphql`,
+      this.apiEndpoint,
```
sprak3000, line 57: Alternatively, if you don't want the main GitHub URL in the xbar variable declaration, you should probably have the default here:

```suggestion
      this.apiEndpoint ?? 'https://api.github.com/graphql',
```

sprak3000 (approved):

alan-nf:
Thanks, @sprak3000 !

PR #2086, opened 2025-02-08
Switched from requests to urllib in LibreLinkUp plugin

skalyan91:
When running the currently-available LibreLinkUp plugin, I kept getting a `ModuleNotFoundError` triggered by `import requests`, probably because Homebrew no longer provides the `python-requests` package. This problem goes away once the API requests are rewritten to use the `urllib` package.

### DIFF
--- a/Lifestyle/librelinkup.1m.py
+++ b/Lifestyle/librelinkup.1m.py
@@ -4,10 +4,10 @@
 
 # <xbar.title>LibreLinkUp Status</xbar.title>
 # <xbar.version>v2.0</xbar.version>
-# <xbar.author>Maurici Abad,Florian Schlund</xbar.author>
-# <xbar.author.github>mauriciabad,FloSchl8</xbar.author.github>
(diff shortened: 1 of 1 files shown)


Lifestyle/librelinkup.1m.py
```diff
@@ -75,78 +75,74 @@ def __init__(self, patient_id, first_name, last_name):
         self.last_name = last_name
 
 headers = {
-    "version": "4.12.0",
+    "version": "4.7.0",
```
sprak3000, line 78: Just curious as to why this version change?

skalyan91: I’m not actually sure, but setting it to anything other than `4.7.0` results in `Error getting patients` (at least on my computer).

sprak3000: @skalyan91 

Is that the same reason the `Accept: application/json` line was removed?

skalyan91: Yes, that’s right.

PR #2106, opened 2025-04-30
multichain compatibile with etherscan V2 endpoint

neila:
- Updated @geeknees's Ethereum gas widget to use the [new V2 endpoints](https://info.etherscan.com/etherscan-api-v2-multichain/).
- Added multichain compatibility by allowing users to input a chain ID of their choice (provided that the chain is supported by etherscan)

### DIFF
--- a/Cryptocurrency/multichain_gas_widget.5s.rb
+++ b/Cryptocurrency/multichain_gas_widget.5s.rb
@@ -4,7 +4,7 @@
 # <bitbar.version>v1.0</bitbar.version>
 # <bitbar.author>Masumi Kawasaki</bitbar.author>
 # <bitbar.author.github>geeknees</bitbar.author.github>
-# <bitbar.desc>Widget for monitoring Ethereum Gas Fees from https://etherscan.io/</bitbar.desc>
+# <bitbar.desc>Widget for monitoring Gas Fees from https://etherscan.io/</bitbar.desc>
 # <bitbar.dependencies>ruby</bitbar.dependencies>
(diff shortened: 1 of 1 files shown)


Cryptocurrency/multichain_gas_widget.5s.rb
```diff
@@ -4,16 +4,17 @@
...
 # <bitbar.abouturl>https://github.com/geeknees/xbar-plugins</bitbar.abouturl>
 
 require 'open-uri'
 require 'json'
 
-ENDPOINT = "https://api.etherscan.io/api?module=gastracker&action=gasoracle&apikey="
-APIKEY = ""
+CHAINID = 1 # Ethereum mainnet. See other available chains at https://api.etherscan.io/v2/chainlist and endpoints at https://forms.blockscan.com/public/grid/3E9QiN00NLhCQVibiP3Z-Bpqhmd7zGXsgapEKJupxiI
+ENDPOINT = "ENDPOINT = https://api.etherscan.io/v2/api?chainid=#{CHAINID}&module=gastracker&action=gasoracle&apikey="
+APIKEY = "" # ETHERSCAN API KEY
```
sprak3000, line 17: While you are updating things, would you consider moving these to be [xbar variables](https://xbarapp.com/docs/2021/03/14/variables-in-xbar.html)?

The `ENDPOINT` could likely remain as a script variable where it needs the chain ID and API key to build it out properly. Also, should it actually be `ENDPOINT = "ENDPOINT = ..."? Feels like a cut / paste error.

PR #2126, opened 2025-10-04
Add plugin for gitlab pipline RAG status

roovo:
Hi there - this plugin adds support for displaying RAG statuses for Gitlab pipelines.

It uses variables for all the user entered configuration (gitlab url, access token, and list of projects to track).

gitlab_pipeline_rag

It's written in ruby.

Many thanks for xbar!


sprak3000:
@roovo 

PR #2130, opened 2025-10-18
Fix AirPodsPower.sh

gonzaloserrano:
Fixes https://github.com/matryer/xbar-plugins/issues/2129


gonzaloserrano:
cc @sprak3000 since I don't know how to add reviewers.

PR #2135, opened 2025-10-27
Add plugin for AWS Cloudformation status

sergiopena:
Plugin to monitor AWS cloudformation stacks status.
It return a list of the stacks matching the configuration variables, info about last update time, last updated status and a link to the AWS console for that stack

### DIFF
--- a/AWS/cloudformation-status.30s.py
+++ b/AWS/cloudformation-status.30s.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env /Users/sergio.pena/.asdf/installs/python/3.13.2/bin/python
+#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
 
 # -*- coding: utf-8 -*-
 


AWS/cloudformation-status.30s.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env /Users/sergio.pena/.asdf/installs/python/3.13.2/bin/python
```
sprak3000: The below ensures python is picked up from either one's path, homebrew, or system installed.

```suggestion
#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
```

sprak3000 (changes requested):

AWS/cloudformation-status.30s.py
sergiopena: @sprak3000 thank you!

I totally forgot I hardcoded that

sprak3000 (approved):

sprak3000:
@sergiopena 

PR #2157, opened 2026-03-20
New plugin to display blood glucose values

elnjensen:
This plugin accesses continuous glucose monitoring data from Dexcom's Share API to display blood glucose values in the menubar.  It can be used either to easily see one's own blood glucose, or to keep track of another person who has given you access to their data.

Screenshot and docs are here:  https://github.com/elnjensen/dexcom_bg_menubar


Lifestyle/Dexcom_BG.1m.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python3
```
sprak3000: This is the standard convention to have the system pick up python from either your path, Apple Silicon based homebrew, Intel based homebrew in that order.

```suggestion
#!/usr/bin/env -S PATH="${PATH}:/opt/homebrew/bin:/usr/local/bin" python3
```

sprak3000 (changes requested):
A suggestion on how you could improve python detection rather than rely on a comment in the code itself.

Also, need to `chmod +x` the script itself.

elnjensen:
Thanks for the suggestions - done.

sprak3000 (approved):

